Issue #12, opened 2017-05-21
items not being removed from the list - fix proposal


ndamjan:
I noticed that when I have 2 browsers open and remove item in one browser, the other one doesn't react to the removal although it gets the notification when using -reactive extension. Same thing happens when re-querying the data and having some deleted records in the meantime between refresh: the list only gets additions and removals are kept.  

So I updated the action addOrUpdateList with the removal logic (see comments). 
Could there be some side-effects of changing this I could not see?

```javascript
  const actions = {
    addOrUpdateList ({ state, commit }, list) {
      let toAdd = []
      let toUpdate = []
      let toRemove = [] // Added

      // Find IDs from the state which are not in the list
      state.ids.forEach(id => {
        if (!list.some(c => c.id === id)) {
          toRemove.push(state.keyedById[id])
        }
      })

      list.forEach(item => {
        let id = item[idField]
        let existingItem = state.keyedById[id]

        checkId(id, item)

        existingItem ? toUpdate.push(item) : toAdd.push(item)
      })

      commit('removeItems', toRemove) // commit removal
      commit('addItems', toAdd)
      commit('updateItems', toUpdate)
    },
```

marshallswain:
Seems like a good fix to me.  Care to start a PR?  I can try to add a test for this.

pmesgari:
Hi, there is still an issue with deleting items through services, the list does not update. I have created a service called `rawsales` with `feathers generate service` and I call the delete method from `feathers-reactive` like this `this.$store.dispatch('rawsales/remove', this.idToRemove)`, when I look into database the record is deleted with `idToRemove` but the `rawsales/list` or the state in the store does not change.

pmesgari:
I found out the issue. It was not with the service but I had defined the `idField` as `_id` in `feathers().configure(rx(RxJS, {idField: '_id'}))` while it must have been `feathers().configure(rx(RxJS, {idField: 'id'}))`

PR #13, opened 2017-05-22
fix item removal in addOrUpdateList (#12)

ndamjan:
As described in #12


marshallswain:
Cool.  I'll add a test.

PR #14, opened 2017-05-25
Bugfix - add params to patch action service call

ndamjan:
Params parameter was not forwarded to the service call from the dispatch method, just added the parameter to the call.


marshallswain:
Thanks!  I'm going to merge this.  I recognize there are problems with the current solution when not using feathers-reactive.  It will take some more work to fix that scenario, later.  However, the primary use case it to use feathers-reactive, with which this fix works well.

marshallswain:
Released as `feathers-vuex@0.4.1`.

PR #20, opened 2017-06-07
QuickFix: Use idField for removal

cmeissl:
The toRemove list in the addOrUpdateList action was built using a hard coded value for the id instead of the idField.


marshallswain:
Thanks.  released as `0.4.2`.

PR #32, opened 2017-08-03
Return Promise.reject for action path of services when catching an er…

phenrigomes:
This commit resolves the issue with patch action. The action does not return a Promise.reject on catch.


marshallswain:
Thanks!  Published as `feathers-vuex@0.7.2`

PR #41, opened 2017-08-23
Fix missing idField

mgesmundo:
The `get` method of the `service-module` `getters` do not pass `idField` to the select function.


marshallswain:
Thanks!  I noticed this around the same time and have pushed the fix.  I've released it as `feathers-vuex@1.0.0-pre.2`.

Issue #43, opened 2017-09-01
Incompatible with latest version of feathers-reactive


saswatds:
This error appears to come only when I start using the feathers-reactive v0.5.1. With v0.4.1 there appears to be no such error

TypeError: Method Promise.prototype.then called on incompatible receiver #<Promise>
    at Promise.then (<anonymous>)
    at Promise.catch (<anonymous>)
    at find (actions.js?1939:38)
    at Array.wrappedActionHandler (vuex.esm.js?358c:598)
    at Store.dispatch (vuex.esm.js?358c:341)
    at Store.boundDispatch [as dispatch] (vuex.esm.js?358c:266)
    at VueComponent.created (RfidReader.vue?c5fe:69)
    at callHook (vue.esm.js?efeb:2556)
    at VueComponent.Vue._init (vue.esm.js?efeb:4000)
    at new VueComponent (vue.esm.js?efeb:4169)

```
if (service.rx) {
      Object.getPrototypeOf(request).catch(handleError) // <-----
} else {
       request.catch(handleError)
}
```

nhim175:
@saswatds it happened to me as well. It was ok a few weeks ago.

marshallswain:
There's a breaking change to `feathers-reactive`, so this plugin is currently only compatible with the previous version.

marshallswain:
`feathers-reactive` is no longer needed with the latest versions of this package. `feathers-vuex` now handles reactivity on its own.  I have no plans to support the newest version of `feathers-reactive`, since it would involve doing work for a feature that already exists without using it.

PR #47, opened 2017-09-29
Update Readme

ImreC:
Reflecting the changes that RxJS does not need to be imported separately anymore. When using the example as a template, it currently throws an app breaking error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,9 @@
 npm install feathers-vuex --save
 ```
 
+## Important
+The current version of `feathers-vuex` is not compatible with the latest version of `feathers-reactive` (0.5.x). To keep on using `feathers-vuex` install version 0.4.x.
+
 ## Use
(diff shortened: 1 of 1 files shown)


marshallswain:
@ImreC thanks for looking into this!  Does this change fix the error described in this issue: https://github.com/feathers-plus/feathers-vuex/issues/43 ?

ImreC:
Not sure. Let me check.

marshallswain:
Looking at the new API, I think it will require some changes:
https://github.com/feathersjs/feathers-reactive#05-observables-are-always-lazy

ImreC:
Yeah, just checked and the error persists

marshallswain:
Thanks for checking, though.  For now, if you want to modify the PR to just add a comment about this plugin only being compatible with `feathers-reactive@0.4.x`, I'll merge it in.

ImreC:
I updated the pull request. I wasn't aware of #43 and am starting a new app, so didn't notice that the actions on the service would break.

marshallswain:
Thanks for the updates.  Looks good.

Issue #48, opened 2017-10-01
setCurrent copy only works with object, not when passing id


plasticrake:
### Steps to reproduce

(First please check that this issue is not already solved as [described
here](https://github.com/feathersjs/feathers/blob/master/.github/contributing.md#report-a-bug))

- [x] Tell us what broke. The more detailed the better.
- [x] If you can, please create a simple example that reproduces the issue and link to a gist, jsbin, repo, etc.

### Expected behavior
When calling mutation `setCurrent(id)` it should copy the current item into `copy` per the README

### Actual behavior
If `id` is passed the copy is set to the `id` instead of the object.

### System configuration

Tell us about the applicable parts of your setup.

**Module versions**
feathers-vuex: 0.8.0

**NodeJS version**: v8.5.0

marshallswain:
@plasticrake that's definitely unexpected, can you try setting a breakpoint at this point in the code and checking why it would be failing: https://github.com/feathers-plus/feathers-vuex/blob/master/src/service-module/mutations.js#L130 ?

My best guess is that you're maybe using MongoDB ObjectIds as the id, instead of the stringified versions of them, so it passes the `isObject` check.

plasticrake:
I created a pull request (#49) based on my understanding of how setCurrent should work. I included a test that shows the issue I'm having.

marshallswain:
I've merged #49.  Thank you!

PR #49, opened 2017-10-01
setCurrent to create copy when passed Id

plasticrake:
### Summary

See #48. 

When calling mutation `setCurrent(id)` it should copy the current item into copy per the README

I included a test that fails before the change made to `setCurrent`. Before if `setCurrent` was passed an Id the `copy` would be set to the Id instead of the object.


marshallswain:
Closes #49

Issue #52, opened 2017-10-10
[1.0] initAuth not exported (or readme is wrong?)


CalvinWalzel:
The readme tells us to import the initAuth function like this:
```
import feathersVuex, { initAuth } from 'feathers-vuex';
```
but apparently initAuth is never exported inside index.js.

Importing it like this works though:
```
import feathersVuex from 'feathers-vuex';
import { initAuth } from 'feathers-vuex/lib/utils';
```

Either the readme should be updated to reflect that, or initAuth should also be exported from the index.js

PR #53, opened 2017-10-10
Export initAuth from index

CalvinWalzel:
Proposal for: #52


marshallswain:
Thanks @CalvinWalzel!!

PR #60, opened 2017-10-13
fix #58 empty error state on action service failure

### LINKED ISSUE #58: errorOn state set to empty object by action error handlers
### Steps to reproduce

Execute an action on the store that results in a rejected promise from the feathers service.

The action error handlers use `Objects.assign({}, error)` to construct the mutation payload which results in an empty object. You can see this in action here, https://runkit.com/embed/4dduppvm6ldg

The mutation methods call `serializeError(payload)` which seems to indicate that is expecting an error type, so the `Objects.assign` can probably be removed.

### Expected behavior
Expect `errorOn*` state (i.e. `errorOnFind`) to have a non-empty Object with properties from the error.

```javascript
errorOnFind: {
  message: 'an error happened'
}
```

### Actual behavior
An empty object is assigned
```javascript
errorOnFind: {}
```
### System configuration

**Module versions** (especially the part that's not working):
feathers-vuex: 1.0.3

**Operating System**:
macOS Sierra

**Browser Version**: 
Chrome 61

cronik:
If an action service method call resulted in an error an empty object
is committed to the `set*Error` mutation rather than the error details.
This was an unintended result of `Object.assign({}, error)` which
always returned an empty object.

The mutation is responsible for serializing the error, via the
‘serialize-error’ package, and updating the `errorOn*` state so passing
the rejected error object as the mutation argument should work.

### DIFF
--- a/src/service-module/actions.js
+++ b/src/service-module/actions.js
@@ -17,7 +17,7 @@ export default function makeServiceActions (service) {
         return response
       }
       const handleError = error => {
-        commit('setFindError', Object.assign({}, error))
+        commit('setFindError', error)
         commit('unsetFindPending')
         return Promise.reject(error)
       }
@@ -57,7 +57,7 @@ export default function makeServiceActions (service) {
           return item
         })
         .catch(error => {
-          commit('setGetError', Object.assign({}, error))
+          commit('setGetError', error)
           commit('unsetGetPending')
           return Promise.reject(error)
         })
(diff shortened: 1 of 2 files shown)


marshallswain:
This is beautiful.  Thank you.

marshallswain:
@cronik please come talk to me in the Feathers Slack when you get a moment.  I'd love to chat.

marshallswain:
Released as `feathers-vuex@1.0.4`.  Thanks again!

PR #65, opened 2017-11-01
Fix default state reactivity

### LINKED ISSUE #64: Auth module reactivity is inconsistent
I've created a simple app with nuxt and feathers to test feathers-vuex and see what I could do with it. I created a git repository here https://github.com/marsimeau/feathers-nuxt-new. I've played with authentication and ran into a problem.

When I dispatch the `authenticate` action for the first time (after a browser refresh) my component won't re-render with the updated payload. In the following example, I still see "No user connected" even though I can see the store has been updated:

```
<template v-if="payload">
  Current user : {{ payload.userId }} <button type="button" @click="logout">Logout</button>
</template>
<template v-else>
  No user connected
</template>
```

If I refresh the page, everything works as it should. I can logout and log back in without facing the issue. The moment I refresh the page while being logged out, the issue comes back.

If feathers-vuex were to be the culprit, I think the problem would be related to the auth store's default values:

```
// feathers-vuex/src/auth-module/state.js
const state = {
  accessToken: undefined, // The JWT
  payload: undefined, // The JWT payload
// ...
```

In my experience with Vue.js and Vuex, in order to be reactive a property needs to be defined. It would explain my issue. Another solution would be to use `Vue.set(state, 'payload', payload)` instead of `state.payload = payload` in mutations (https://vuex.vuejs.org/en/mutations.html#mutations-follow-vues-reactivity-rules) here:

```
// feathers-vuex/src/auth-module/mutations.js
setPayload (state, payload) {
  state.payload = payload
},
```

Doing this would ensure reactivity for any of the state items. After a bit of testing I cannot find any other cause to this issue.

marsimeau:
This PR would fix #64 .

I've changed every default state that was undefined for null.


marshallswain:
Thank you!

marshallswain:
Published as `feathers-vuex@1.1.1`  <--- We can drink to that.  It's happy hour!

PR #66, opened 2017-11-01
Clarify api change inline

rayfoss:
Should be clear, even without coffee.

It's easy to forget patch and update don't use the same API as feathers.

Added an example use that looks more like what people should be doing in a component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,7 +292,7 @@ store.dispatch('todos/create', newTodo)
 
 #### `update(paramArray)`
 Update (overwrite) a record.
-- `paramArray {Array}` - array contailning the three parameters update takes.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -303,11 +304,26 @@ let params = {}
...
+  }
+}
+```
+
+#### `patch(paramArray)`
 Patch (merge in changes) one or more records
-- `id {Number|String}` - the `id` of the existing record being requested from the API server.
-- `data {Object}` - the data that will be merged into the existing record
-- `params {Object}` - An object containing a `query` object.
+- `paramArray {Array}` - array contailning the three parameters patch takes.
```
lkchoi-lucyio: typo: containing

marshallswain:
Thank you!

PR #73, opened 2017-11-11
Added auth 'userService' option example in README.md

Cl3MM:
I had a hard time finding the 'userService' option in the documentation, I actually had to read the source code to find it out, so it made sense to me to add an example.


marshallswain:
Beautiful!  Thank you!

PR #95, opened 2018-03-02
Use serialize-error for auth errors mutations

louim:
Hey!

This is kinda related to https://github.com/feathers-plus/feathers-vuex/pull/60. Errors for the `auth` module where not set correctly. I used the same logic from the service module mutations.


marshallswain:
Ah.  Very nice.  Thanks!

PR #102, opened 2018-03-16
auth setting base on custom route

rofiuddin15:
How an authentication setting if the route of service or rest like this [ /api/service ]
I have problem with it, I can't get the user value
my code on store like below
```
plugins: [
    service('/api/users'),

    auth({userService: '/api/users'})
  ]
```


marshallswain:
That is supposed to work.  What are you experiencing?  I'm going to close this PR because it seems more like an issue.

PR #107, opened 2018-03-31
Remove semicolons to stay consistent

FreeLineTM:
I went crazy from the inconsistent use of semicolons from line to line.


marshallswain:
Thanks!  Die, semicolons!

Issue #110, opened 2018-04-13
enhance the `updated/patched` event handling


beeplin:
### Steps to reproduce

(First please check that this issue is not already solved as [described
here](https://github.com/feathersjs/feathers/blob/master/.github/contributing.md#report-a-bug))

### Expected behavior

Suppose we have a service-module `A` which is initially empty. Some one else updated an item with _id=’a' to the server, so we received an `updated` event. The ideal way to handle this event is to have an options to control:

1. if we want to put this new item into state, then call `commit('A/addItem', item)`；
2. if we want to discard this new item which is not previously in the state, then do nothing.

Since in feathers v3 we are able to efficiently filter the event publications in the back-end via `channels`, I guess in most cases people want to keep new pushed items in their states (as in the first case)

### Actual behavior

Currently the 'updated` event is handled by:

https://github.com/feathers-plus/feathers-vuex/blob/master/src/service-module/service-module.js#L66

which calls:

https://github.com/feathers-plus/feathers-vuex/blob/master/src/service-module/mutations.js#L25

Since the module is previously empty, `state.keyedById[id] = item` is actually adding a new property `a` to `state.keyedById`. According to https://vuejs.org/v2/guide/list.html#Object-Change-Detection-Caveats, this new property is not reactive and hence can not be seen in components/DOM, although it is occupying browser's memory.

This means the new items is taking browser resourse but cannot be used at all.

### System configuration

Tell us about the applicable parts of your setup.

**Module versions** (especially the part that's not working):

**NodeJS version**:

**Operating System**:

**Browser Version**:

**React Native Version**:

**Module Loader**:

PR #111, opened 2018-04-13
Add an 'upsert' option to service modules

beeplin:
As discussed in #110, this 'upsert' option allows users to choose whether they want to discard the new (currently not existing in the store) records pushed by 'updated/patched' socketio events or to add them into store.

Of coz we can discuss for a better name. The term 'upsert' is used in some database to indicate the similar behavior.


beeplin:
closed via https://github.com/feathers-plus/feathers-vuex/issues/42#issuecomment-381308174

Issue #42, opened 2017-08-24
[discussion] using feathers-vuex as a client-side incremental cache


beeplin:
One of the wonderful things about `feathers-vuex` is that it loads back-end data, keeps them in browser's memory (vuex states), and updates them via socketio when needed. This makes it a handy tool to do client-side cache so that we don't need to waste band-width to request the same data again and again.

Here I want to discuss what we should do to make it easier to turn feathers-vuex into an out-of-box client-side incremental cache. 

Previously I use raw client-side feathers services to cache data, and use `feathers-reactive` and `vue-rx` to let vue components be able to subscribe these data from the client-side feathers services. This approach requires considerable knowledge of `rxJS`, and not all team members feel comfortable with it.

Another possibility is to use the newly developed `feathers-offline-realtime` and `feathers-offline-publications`. Pros: they provide out-of-box data-replica and syncing between client and server, and furthermore, allow easy configuration and optimization of event filtering on the the server side. Cons: 1) they are designed to sync the whole back-end service data or a subset of the data with a given query, so they do not support incremental cache (i.e. querying and caching some records at first, and then run new `find\get` methods to add more records into the cache later); 2) Still need to write some rxJS code to glue feathers data and vue components.

I have discussed similar topic with @eddyystop on the `feathers-offline-realtime` project (https://github.com/feathersjs/feathers-authentication-management/issues/8#issuecomment-313148338). Later I realized the `realtime` strategy, by its nature, is for *replication* of a given service over a given query, not for incremental caching.

I think `feathers-vuex` is currently the best way to fulfill a full-functional client-side incremental cache for applications using feathers and vue. It naturally allows incrementally adding new records into the store, and since it uses vuex, no need for rxJS subscriptions. 

Based on what we already have, I propose several new APIs for the `service` module, which can make the caching process easier:

1) a new option in the `params` object  for the `get` action, which makes the `get` firstly look for the given id in the vuex state (similar to the `get` gettter). If the item exists, bypass the remote request and just return the item in the vuex state with a resolved promise; if not, request the item from remote server normally, cache it in the vuex store, and return it (similar to the original `get` action).

2) a new option in the `params` object for the `update` and `patch` actions,  which enables optimistic mutation, i.e., mutating the data in the vuex state first, letting the back-end do the same mutation, and in case of back-end error, restoring the old data to the vuex state. Currently we can write some code to use the `copy` object to do this job, but it would be nice to include this in the offical API.

[...]

PR #112, opened 2018-04-14
Improve get action & updateItem mutation

beeplin:
Add option `skipRequestIfExists` for get action: if the record already exists in store, skip the remote request. #42

Add option `upsert` for enhanced updateItem mutation: add new records pushed by 'updated/patched' socketio events into store, instead of discarding them. #110


marshallswain:
@beeplin this looks beautiful.  I feel like the default for `upsert` should be `true`, since people would expect it to be created, anyway.  Do you have any thoughts against it?

marshallswain:
What do you feel about 

`upsert` —> `addOnUpsert`

`skipRequestIfExists` —> `localOnly`

???

marshallswain:
Also, let’s definitely set upsert to true by default. This is technically a bug fix, since the code was intended to work this way, but I messed up the mutation.

beeplin:
`localOnly` might be a little bit confusing, because if the record is not in state, there will still be a remote request, which does not sound like `localOnly` literally.

`addOnUpsert` sounds great~

marshallswain:
@beeplin Please let me know when this is ready to merge.

beeplin:
Sorry has  been busy this week. only 2 things left for this PR: rename the two new params, and write tests for the new get actions (yet to learn how).

I am afraid I would have no enough spare time to complete it until next Wenesday.  Is it troubling your schedule for the new release?

marshallswain:
Everything is fine. I’m working on documentation for a while, anyway.
Thanks!

marshallswain:
Released as part of `feathers-vuex@1.2.0`

PR #119, opened 2018-05-09
Dates in instanceDefaults

marshallswain:
You can now pass the Date class as the value of an instanceDefault to have that value automatically converted to a Date.

```js
new Vuex.Store({
  strict: true,
  plugins: [
    service('todos', {
      instanceDefaults: {
        id: null,
        description: '',
        isComplete: false,
        createdAt: Date
      }
    })
  ]
})

it('converts keys that contain the Date constructor into date instances', function () {
  const { Todo } = this
  const createdAt = '2018-05-01T04:42:24.136Z'
  const todo = new Todo({
    description: 'Go on a date.',
    isComplete: true,
    createdAt
  })

  assert(typeof todo.createdAt === 'object', 'todo.createdAt is an instance of object')
  assert(todo.createdAt.constructor.name === 'Date', 'todo.createdAt is an instance of date')
  assert(todo.createdAt.toString() === new Date(createdAt).toString(), 'the correct date was used')
})
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "feathers-vuex",
   "description": "FeathersJS, Vue, and Nuxt for the artisan developer",
-  "version": "1.2.1",
+  "version": "1.3.0",
   "homepage": "https://github.com/feathersjs/feathers-vuex",
   "main": "lib/",
   "keywords": [
--- a/src/service-module/model.js
+++ b/src/service-module/model.js
@@ -25,6 +25,15 @@ export default function (options) {
           // Reset the instance default for this prop to null
           _instanceDefaults[key] = null
         }
+
+        // Or if the value is a Date
+        if (modelName === Date) {
+          // Store the relationships
+          relationships[key] = Date
+
+          // Reset the instance default for this prop to null
+          _instanceDefaults[key] = null
+        }
       })
 
       if (options.isClone) {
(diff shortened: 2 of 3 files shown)


pbpraha:
Hello,

First, thanks for a great project, I'm really loving the realtime aspect.

Regarding this feature:
I've a 'Todo' record, it has a 'due' date field.
I have the service set up and running fine, i can see data in the Vuetify datatable.
however, if a record is ever patched or saved then the 'due' field is turned into a String, and as such can no longer be queried with e.g. {$gt:[somedate]}

What am I doing wrong here, or is this a bug ?

marshallswain:
When the data is serialized on the client (in preparation for sending it to the API) it converts it to a string.  Your API server will need to convert it back to a date object before saving it again.

pbpraha:
Hello Marshall,

thanks for the quick answer, unfortunately it does not answer this issue.

the patch operation does not update the entire record, only one field (that is not a date field)

this is not an issue at the back end. I've watched (via chrome dev tools) the websocket. the data provided to the app initially and in the response from the server after update is identical - and as you say, is serialised and dates are strings.

my app is correctly casting (via instanceDefaults) the field as a Date when the initial 'fetch all the data from the server' query runs but does not then cast the field as a date when subsequent record updates are received on the socket.

this is a deal breaker and i'm wondering what i'm doing wrong...

marshallswain:
Where is the data not queryable?  In the API server, or locally with the feathers-vuex getters?

pbpraha:
Locally.
i've tried using both vanilla feathers-vuex getters and feathers-vuex-find components and the result is the same in both instances

marshallswain:
Ok. I misunderstood where the issue was.  I'll switch the conversation over to the issue you created.

https://github.com/feathers-plus/feathers-vuex/issues/228

pbpraha:
Thanks :)
sorry for not being clear here, I had a hard time finding the right place to put this issue.

PR #126, opened 2018-05-24
Update pagination to use Vue.set to ensure it is reactive

Mattchewone:
### Summary
In ref to https://github.com/feathers-plus/feathers-vuex/issues/125

This is set to use `Vue.set` to update the pagination object so that it is kept reactive to updates.


marshallswain:
Released as `feathers-vuex@1.4.1`.

PR #139, opened 2018-07-10
Change constructor.name to constructor.className

ucokfm:
Fix [https://github.com/feathers-plus/feathers-vuex/issues/136](https://github.com/feathers-plus/feathers-vuex/issues/136)

Instead just removing the existing codes, i found we can use another property name to hold the new name and modify tests to use that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "feathers-vuex",
-  "version": "1.4.5",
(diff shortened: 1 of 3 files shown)


marshallswain:
Thanks for this!  I'll do a release shortly.

marshallswain:
Published as v1.4.8.  Thanks!

Issue #142, opened 2018-07-25
replaceItems appears to break cloning of (replaced) item


jscottsf:
### Steps to reproduce

1. Turn on replaceItems

2. Use instance.clone() on a given item then commit. The item will be replaced as expected.

3. Given the item was bound to a prop, then try to do clone again. You'll get `.clone is not a function`.

Upon further inspection, you'll see that the first clone is a FeathersVuexModel (with isFeathersVuexInstance: true), whereas the second clone is not, even though it has the model properties and is observed.

On a replace item, would it be better to simply remove all of the item's properties, and replace them with the new item properties? Reassigning the object seems to break a lot.

jscottsf:
Upon deeper digging, it seems that socket events invoke the mutation `updateItem`. This is normal. However, `updateItem` does not assign a replaced item as a Model. Hence it looses the ability to be cloned.

```
service.on('updated', item => store.commit(`${namespace}/updateItem`, item))
```

No Model is created in `updateItem`. Shouldn't this be similar to `addItem`?

```
    // Simply rewrite the record if the it's already in the `ids` list.
    if (isIdOk && state.ids.includes(id)) {
      if (replaceItems) {
        state.keyedById[id] = item
      } else {
        _merge(state.keyedById[id], item)
      }
      return
    }
```

marshallswain:
Closing, since the fix has been merged.  Thank you!

PR #144, opened 2018-07-30
Replace items should assign state as a Model

jscottsf:
### Summary

This a fix for issue #142 .

Tested and works.


marshallswain:
Makes sense.  Thank you!

PR #167, opened 2018-09-24
Fix slowness when finding a large number of items

ricochet1k:
addOrUpdateList was duplicating work, and the keyedById object
was being cloned, handed to Vue to be made reactive before being thrown away,
over and over again through the loop. Instead, now do as much work as possible
by batching addItem mutations together.


marshallswain:
Thank you for this!

marshallswain:
Released in `feathers-vuex@1.6.0`.

Issue #179, opened 2018-11-14
Add toJSON method to FeathersVuexModel class


NickBolles:
### Steps to reproduce
- [ ] Tell us what broke. The more detailed the better.
- [ ] If you can, please create a simple example that reproduces the issue and link to a gist, jsbin, repo, etc.

I'm not exactly sure what is causing this, besides that I'm doing SS authentication, and trying to store the 
```
‼ warn Cannot stringify arbitrary non-POJOs FeathersVuexModel
```

I think this could be resolved by adding a `toJSON` method on the class. Plus in-general it's a good practice. 

https://github.com/nuxt-community/devalue/pull/1

DreaMinder:
So is this resolved by adding toJSON? Because I still got warnings
`"feathers-vuex": "^1.6.2",`

NickBolles:
@DreaMinder I'm not sure that there's been another release to non since this change
@marshallswain do you know?

marshallswain:
Lots of them in the `pre` channel

`npm i feathers-vuex@pre`

PR #197, opened 2019-02-15
Use merge for toJson object

### LINKED ISSUE #196: Basic usage breaks in Nuxt SSR
FeathersVuex breaks on Nuxt SSR with the most basic usage.

`$FeathersVuex.Todo` items load fine on client side (open root, then click the link to `/todos`);

On server side (refresh `/todos`), infinite recursion happens in `toJSON` in `feathers-vuex/lib/service-module/model.js:237`.

### Steps to reproduce

See this repo/branch for a reproduction:
https://github.com/Palivonas/feathers-vuex-bug-reproduction/tree/models-tojson-nuxt-ssr

FeathersVuex `1.7.0-pre.54`.
Node 10 and 8 tested, app created using `create-nuxt-app`.

NickBolles:
This will avoid infinite recursion with circular references

### Summary
#196 - infinite recursion when there are circular references
closes #196, caused by #179 
caused by #179 

#### Question
- Is merge the best route here? it's probably the easiest, but maybe not the most performant  if we already know the key we want to copy to the json object.
- Should we use merge for `cloneWithAccessors`?

### DIFF
--- a/src/service-module/model.js
+++ b/src/service-module/model.js
@@ -1,4 +1,5 @@
 import fastCopy from 'fast-copy'
+import _merge from 'lodash.merge'
 
 const defaults = {
   idField: 'id',
@@ -170,7 +171,7 @@ export default function (options) {
     _remove () {}
 
     toJSON () {
-      return fastCopy(this)
+      return _merge({}, this)
     }
   }
 


NickBolles:
@marshallswain could you take a look at my questions in this? 

@Palivonas this passes your reproduction repo, are there any other test cases to do?

marshallswain:
I’ll bet this won’t work with getters whicj have no setter.

NickBolles:
@marshallswain if they are enumerable it should work, check this code sample, the last output is the most relevant

https://runkit.com/nickbolles/lodash-merge-and-defineproperties

marshallswain:
@NickBolles It seems to me like lodash.merge works exactly the same as copyWithAccessors.  Thanks for setting up that runkit.

marshallswain:
Published as `feathers-vuex@1.7.0-pre.56`

NickBolles:
@marshallswain NP! 

So do we need to add any documentation to the instance defaults (or anything else) to make sure that getters are enumerable if you want the property serialized to JSON (such as when SSR and sending it to the client). #196 is a good example that we could point to the documentation.

marshallswain:
@NickBolles Yeah.  Feel free to make a note in the /docs repo.

PR #208, opened 2019-03-01
Fix defineProperties in instanceDefaults

### LINKED ISSUE #207: Model.get() breaks with instanceDefaults and defined properties
Example service plugin with a getter in Vuex store:
```js
service('events', {
  instanceDefaults(data, { Models }) {
    return Object.defineProperties({}, {
      location: {
        get() {
          return Models.Location.getFromStore(this.locationId);
        },
      },
    });
  },
});
```

Using Object.defineProperties as discussed in https://github.com/feathers-plus/feathers-vuex/issues/196#issuecomment-464484399 to make sure it doesn't break toJSON and is not included in patch/update requests.

Calling `Models.Event.get('someId')` works as expected if `someId` does not exist in the store. If it is already in the store, a TypeError is thrown:
```
Cannot redefine property: location
    at Function.defineProperty (<anonymous>)
    at model.js:133
    at Array.forEach (<anonymous>)
    at new FeathersVuexModel (model.js:131)
    at Store.setCurrent (mutations.js:262)
    at wrappedMutationHandler (vuex.esm.js:714)
    at commitIterator (vuex.esm.js:382)
    at Array.forEach (<anonymous>)
    at vuex.esm.js:381
    at Store._withCommit (vuex.esm.js:512)
```

Using feathers-vuex@1.7.0

Palivonas:
Skip redefining getters, setters and functions

Fixes #207

### DIFF
--- a/src/service-module/model.js
+++ b/src/service-module/model.js
@@ -43,7 +43,7 @@ export default function (options) {
       Object.keys(_instanceDefaults).forEach(key => {
         // Prevent getters and setters from firing before the instance is constructed
         const desc = Object.getOwnPropertyDescriptor(_instanceDefaults, key)
-        if (desc.get || desc.set) {
+        if (desc.get || desc.set || typeof desc.value === 'function') {
           return
         }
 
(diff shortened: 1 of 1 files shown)


Palivonas:
Though I'm wondering why `service/addOrUpdate` action does not have the same issue. Shouldn't a constructor use the same code to add or update records? If so, would it make sense to extract that code and reuse in both `addOrUpdate` and model constructor?

Palivonas:
As mentioned in #207, defining instanceDefaults properties with `configurable: true` fixes the issue without updating feathers-vuex source code. That does seem like a bit of a workaround though. Any thoughts?

marshallswain:
Oh thank you so much for this!  This was on the list of issues to fix in my head.  In answer to your question about making the action perform the same thing, I remember I had a reason that I didn't make them the same, but now I can't remember it.  It may not even be something that's relevant at this point.  I agree that there's a bit of cleanup work to do, though.

PR #212, opened 2019-03-21
Patch 1

fratzinger:
### Documentation Changes

I found some inconsistencies between docs and code. So I added some states and mutations in the docs, which where missing and described somewhere else (in the code oder other examples) anyway. I changed the order of the states und mutations in the code to match the order in the docs for less confusion.

Thanks for this awesome piece of code! It's really enjoyable to program with :heart:


marshallswain:
Thanks!

Issue #224, opened 2019-05-13
instanceDefaults as an object can cause complications


james2mid:
**Neither a bug or a feature request**

I'm posting this for myself to come back to at a later time unless someone else wants to do it.

The issue is with instanceDefaults as an object where it maintains the contents of any array or object properties between model instances. The workaround for this is to make instanceDefaults a function that returns the same object.

It took me a little while to figure out why this was happening and so I felt this should be more clear in the docs, especially for the less versed coders out there.

james2mid:
The code below will have `arr` maintain its value between instances when doing `new this.$FeathersVuex.Model()`.

```js
instanceDefaults: {
  arr: []
}
```

The workaround is to do this:
```js
instanceDefaults: () => ({
  arr: []
})
```

marshallswain:
@james2mid thanks for making this more visible.  Version 2.x will only allow the function version.  I'm going to close this after it is released.

james2mid:
See the PR above

PR #227, opened 2019-06-10
Warn against instanceDefaults as an object #224

james2mid:
# Documentation change only

Discussed in issue #224


marshallswain:
Thanks for this!

PR #231, opened 2019-06-28
Adding dynamically generated QID in the MakeFindMixin

guzz:
I've been strugling to get the pagination to work correctly and one of the problems was that I couldn't create dinamical qid's, so I would have the comments of a post show a total of the pagination of a different post that was visited before as they would share the same qid.

This is a simple implementation and maybe there is something already been done for the next version, but anyway.

### DIFF
--- a/docs/mixins.md
+++ b/docs/mixins.md
@@ -76,7 +76,7 @@ The `makeFindMixin` and `makeGetMixin` utilities share the following options in 
 
 The `makeFindMixin` has these unique options:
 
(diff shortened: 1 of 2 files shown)


marshallswain:
Thanks @guzz I need to inspect how well this works with the code in the pagination branch.  Pagination has been completely reworked in that branch.

guzz:
For what I looked, it needs some changes but I'm not able to test it, I'm getting an error with steal.js in the pagination branch. I'll try to figure it out latter today.

marshallswain:
@guzz I’ve manually merged the pagination branch into the typescript branch.  That branch will soon become 2.0.  You can try your code out in that branch probably a bit easier.  Use the `dist` folder installed from npm and it should work better with Steal.

guzz:
Ok, @marshallswain but reading the pagination branch I found the way you are doing this dinamic qid is to put it in the queryParams, and I think this would be redundant, and also was giving some trouble to manage not to break the possibility to declare it in the queryParams.

PR #258, opened 2019-09-05
addItems mutation fix tempId

fratzinger:
Currently I'm programming an progressive web app with feathers-vuex and localForage. I stumbled upon this. I wanted to restore `keyedById` from IndexedDB by using `addItems`. Instead of just overwriting the stores `keyedById` by the IndexedDB ones, I wanted to have them as Vuex-Models :)

This little change should fix the problem. Everything works for me now.

If `tempId` was assigned before calling `addItems`-mutation -> `tempId` is undefined -> `newTempsById[undefined] = item`

Again thanks for this great module! <3

### DIFF
--- a/src/service-module/service-module.mutations.ts
+++ b/src/service-module/service-module.mutations.ts
@@ -41,7 +41,7 @@ export default function makeServiceMutations() {
 
       if (isTemp) {
         let tempId = item[tempIdField];
-        if (!tempId) {
+        if (tempId == null) {
           tempId = assignTempId(state, item)
         }
         item.__isTemp = true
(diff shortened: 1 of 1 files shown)


src/service-module/service-module.mutations.ts
```diff
@@ -40,8 +40,9 @@ export default function makeServiceMutations() {
       }
 
       if (isTemp) {
-        if (!item[tempIdField]) {
-          var tempId = assignTempId(state, item)
+        let tempId = item[tempIdField];
+        if (!tempId) {
```
marshallswain: Please change this to `if(tempId == null) {` so it will work with falsey, but valid, ids.

marshallswain:
Thanks for the great catch.  I've made one comment.  If you could update that tempId `if` statement, I'll gladly merge.

fratzinger:
Didn't think of that. Changed it. Thank you

marshallswain:
Published as `feathers-vuex@2.0.0-pre.76`

PR #261, opened 2019-09-09
Update FeathersVuexFind.ts - pagination

fratzinger:
I think adding queryInfo & pageInfo to the scope of renderless components is a consistent step. I wanted to implement serverside pagination into renderless components and I found no example for that (is there one which I didn't see?). I also wanted to have the values for `total`, `$limit` and `$skip` to show in a pagination component.
So I thought I need the `ids` of pageInfo to pass it to the `query`-prop for vuex. But without knowing the `queryId` & `pageId` I have to stringify the ids by myself.

I hope it got clear and makes sense. Am I trying to use renderless components the right way or should I start using mixins? I don't like mixins that much because it's much more irritating where properties come from.

As always thanks for this peace of code and I'm open for discussion if that doesn't make sense to you :)

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -315,6 +315,65 @@ export default {
 </script>
 ```
 
+### server-side pagination
+
+When you want to use server-side pagination you need to pass the ids from the server to vuex. It can be done by a combination of `query`, `fetchQuery` and `editScope` as described below. The `fetchQuery`-prop is only computed after items from the server arrived. The ids for the `find` getter as well as the total amount of available values `total` are extracted by the `edit-scope` function and stored in `data`:
+
+```html
+<template>
+  <FeathersVuexFind
+    :service="service"
+    :query="internalQuery"
+    :fetch-query="fetchQuery"
+    :edit-scope="getPaginationInfo"
+  >
+    <div slot-scope="{ items: todos }">
+      {{todos}}
+    </div>
+  </FeathersVuexFind>
+</template>
+
+<script>
+export default {
+  data() {
+    return {
+      service: 'users',
+      ids: [],
+      query: {
+        isComplete: true
+      },
+      total: 0,
+      limit: 10,
+      skip: 0
+    };
+  },
+  computed: {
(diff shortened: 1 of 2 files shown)


marshallswain:
This looks well implemented.  Do you think you could address the errors on Travis CI?

marshallswain:
@fratzinger please let me know if you need me to fix this from my end.

fratzinger:
nono, fixes are clear. I'm good. But I figured out how to implement server-side pagination with this and now I'm adding it to the docs. Will push it immediately.

marshallswain:
Awesome.  I really appreciate quality PRs like this.  I implemented this stuff into the mixins, so it's great to have the feature parity.  It makes 2.0 even better than I planned.  Thank you very much.

fratzinger:
Thanks. Can only return the compliments. I've added an example. It works for me, but I don't know if it's the best way to do it.
I figured out that `editScope` gets called 5 times in my server-side pagination example. Do you have an idea why? Is this exactly the amount of properties of the scope? Is this something to worry about or maybe changable?

I also asked myself if server-side pagination should be an boolean-prop for the component so you only have to pass a `query`. `fetchQuery` would be ignored and `editScope` is not needed. What do you think @marshallswain ?

marshallswain:
I feel like the way you have it right now is perfect.  The boolean-prop idea does provide a simple solution for one use case, but it limits others.  I'm going to merge and release this, now.

marshallswain:
Released as `feathers-vuex@2.0.0-pre.77` . Thank you!

Barbapapazes:
@fratzinger I'm sorry but when I try your solution, it seems it doesn't work. On server side, pagination is limited at 10 but I have more record than 10 in my database. So I want to be able to load more data using $skip and $limit but I found no way to do this.

Because, your solution load only 10 items, as shown in your example total is set to 10, and the watch properties doesn't work with $skip or $limit. Or maybe, there is something I don't understand ?

PR #347, opened 2019-11-10
Define path to TypeScript types in package.json

JorgenVatle:
This pull request enables TypeScript declarations and adds a reference to them in `package.json`. It essentially prevents seeing a warning when attempting to pull in the package from a TypeScript file. This also enables the TypeScript type hinting we all have come to love. 🎉

image


marshallswain:
Lovely!  Thank you!

PR #365, opened 2019-11-22
remove @ts-ignores

jeffborg:
This overcomes the issue with the Pagination<> not having a generic type. But I also removed all the @ts-ignore tags in the source code.

I have left the test code alone


marshallswain:
Thanks!  That's much nicer!

PR #400, opened 2019-12-26
docs: fix description and example of BaseModel.setupInstance()

jntesteves:
### Summary

Just a simple fix to the doc of this method, since it was still a copy of `instanceDefaults`'s doc.

I've tried to describe what I understood from looking at the code of BaseModel, but I'm not sure I got it right. Feedback will be appreciated.

### Other Information

None.


marshallswain:
Good catch!  Thanks

PR #403, opened 2020-01-06
test(package.json): Fix glob pattern in mocha test script

J3m5:
The test script in the package.json should have the glob pattern wrapped in single quotes like mentioned here: [https://mochajs.org/#the-test-directory](https://mochajs.org/#the-test-directory)

> You should always quote your globs in npm scripts. If you use double quotes, it's the shell on UNIX that will expand the glob. On the other hand, if you use single quotes, the node-glob module will handle its expansion.

If not, in some environments it will not take the files in the root directory into accounts.


J3m5:
If you look at the travis logs, you will see that it picked up the makeFindMixin tests, that are located at the root of the test directory.
If you compare it to the previous PR logs, it's not there.

marshallswain:
Thanks!

PR #407, opened 2020-01-11
improvement(mixin): keep the find method available even when the local option is set, fix #394, 

### LINKED ISSUE #394: [Bug] MakeFindMixin: local option doesn't works as specified in the docs
### Steps to reproduce
In a template, add a makeFindMixin like this: 

```js
mixins:[ makeFindMixin({ service: "yourServiceName", local: true, params: () => ({}) }) ]
```

### Expected behavior
The mixin should not make a request to the API server but use the params to pull data from the local store as stated in the docs.

> local {Boolean|String|Function} - when true, will only use the params prop to pull data from the local Vuex store. It will disable queries to the API server. The value of local will override queryWhen. Default:false

### Actual behavior
The API request is sent, the local option is not respected.
If you omit the params option, the API request is not sent but the store getter returns nothing.

### What cause this behavior
In the created lifeCycle event, on [this line](https://github.com/feathersjs-ecosystem/feathers-vuex/blob/98b3f796050340e8afdeba4183a9685367cd8fbb/src/make-find-mixin.ts#L245) , only
the presence of the params and fetchParams options are checked in order to make the API request.

If they are not present, the absence of the local option is used to log a message.

### How to fix it
Check the truthiness of the local option to send the API request or not.
I can submit a PR if everything is ok.

### System configuration

**Module versions**: 3.2.1

**NodeJS version**: 12.4.0

**Operating System**: Linux

**Browser Version**: Chromium 79

**Module Loader**: Webpack

J3m5:
This pull request fix #394 

To resume, in the mixins, the local option make the mixin not fetch records on the created hook but still gives access to the getter and items.

But it also prevent the use of the find method.
So in order to make a request you need to use the vuex store directly.

This PR lets you set the mixin as local but still have access to the find mixin for a later use.
In addition it makes the mixin a bit more lightweight as the entire created hook is added conditionally.

In order to not introduce a breaking change, the watch option is still ineffective when the local option is set.

In the future, it might be interesting to modify this behavior by letting the watch option works even if the local option is set.
Thus, the local option could be renamed to something like "fetchOnCreated".

This PR also fix [the test for the dynamic service name](https://github.com/feathersjs-ecosystem/feathers-vuex/blob/14c8305acbb182d5ce52eb9b8b75ac38964360d2/test/make-find-mixin.test.ts#L99) that was previously skipped.

### DIFF
--- a/src/make-find-mixin.ts
+++ b/src/make-find-mixin.ts
@@ -4,14 +4,15 @@ no-console: 0,
 @typescript-eslint/explicit-function-return-type: 0,
 @typescript-eslint/no-explicit-any: 0
 */
-import {
-  getServicePrefix,
-  getServiceCapitalization,
-  getQueryInfo,
-  getItemsFromQueryInfo
-} from './utils'
+
 import debounce from 'lodash/debounce'
 import _get from 'lodash/get'
+import {
+  getItemsFromQueryInfo,
+  getQueryInfo,
+  getServiceCapitalization,
+  getServicePrefix
+} from './utils'
 
 export default function makeFindMixin(options) {
   const {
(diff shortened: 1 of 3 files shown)


marshallswain:
Ah, lovely!  I see this as a great performance update and lots of cleanup.  Thanks for fixing the test, too!

PR #421, opened 2020-02-06
Allow simple data modification in clone

mrfrase3:
### Tell us about the problem your pull request is solving.
This allows a simple one-liner `user.clone({ enabled: !user.enabled }).save()`

### Are there any open issues that are related to this?
nope

### Is this PR dependent on PRs in other repos?
nope

### Should this go in the docs
Maybe, this won't account for errors and is probably not the recommended way to go about changes. But it would be a nice hack for power users ;)

### DIFF
--- a/src/service-module/make-base-model.ts
+++ b/src/service-module/make-base-model.ts
@@ -266,7 +266,7 @@ export default function makeBaseModel(options: FeathersVuexOptions) {
       return this._clone(id, data)
     }
 
-    private _clone(id, data) {
+    private _clone(id, data = {}) {
       const { store, namespace, _commit, _getters } = this
         .constructor as typeof BaseModel
       const { keepCopiesInStore } = store.state[namespace]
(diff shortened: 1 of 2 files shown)


marshallswain:
Oh hell yes!  Kill that boilerplate.

marshallswain:
@mrfrase3 think you could add a test?  This is a badass feature that deserves to be protected by a test.  You can probably just duplicate one of the clone and commit tests.

I think this also allows one-line eager updating by throwing a `commit()` in there before `.save()`.

```js
user.clone({ enabled: !user.enabled }).commit().save()
```

I'm really happy with this idea.  I hate those four-line clone and commits.  ;)

src/service-module/make-base-model.ts
```diff
@@ -256,28 +256,28 @@ export default function makeBaseModel(options: FeathersVuexOptions) {
...
         throw new Error('You cannot clone a copy')
       }
       const id =
         getId(this, idField) != null ? getId(this, idField) : this[tempIdField]
-      return this._clone(id)
+      return this._clone(id, data)
     }
 
-    private _clone(id) {
+    private _clone(id, data) {
```
J3m5: private _clone(id, data= {}) {

src/service-module/make-base-model.ts
```diff
@@ -256,28 +256,28 @@ export default function makeBaseModel(options: FeathersVuexOptions) {
...
       const { store, namespace, _commit, _getters } = this
         .constructor as typeof BaseModel
       const { keepCopiesInStore } = store.state[namespace]
 
       _commit.call(this.constructor, `createCopy`, id)
 
       if (keepCopiesInStore) {
-        return _getters.call(this.constructor, 'getCopyById', id)
+        return Object.assign(_getters.call(this.constructor, 'getCopyById', id), data || {})
```
J3m5: return Object.assign(_getters.call(this.constructor, 'getCopyById', id), data)

J3m5:
Just trying something... 
A default parameter can be used to avoid ` || {} `.
Seems cleaner to me.

marshallswain:
Looks great!  Thanks!!!

PR #424, opened 2020-02-11
update link to vue-use-web

morphatic:
The link you had here no longer seems to work 🙁 

New maintainer maybe? Anyway, it's an amazing package, and I wouldn't want people to think it was just gone.


marshallswain:
Thanks for this!

PR #425, opened 2020-02-11
docs: install peer dependency `@vue/composition-api`

jntesteves:
### Install peer dependency on getting-started

When installing/updating to 3.4.1 and up, we get the error `Module not found: Error: Can't resolve '@vue/composition-api'`.
Just adding this new dependency to the install scripts in getting-started docs.


marshallswain:
Thanks for this. I didn't realize that you had added this.

PR #437, opened 2020-03-01
Added check for getterParams is not undefined, to enable documented "…

anli-xsigns:
…return null" feature

Added check for getterParams is not undefined, to enable documented "return null" feature as described in https://vuex.feathersjs.com/composition-api.html#options (before returning null in the computed threw an error because paginate couldn't be found on a null object)


marshallswain:
Good catch.  Thank you!

PR #442, opened 2020-03-07
Update Full nuxt example in common-patterns.md to fix middleware redirect

andrewharvey:
### Summary

(If you have not already please refer to the contributing guideline as [described
here](https://github.com/feathersjs/feathers/blob/master/.github/contributing.md#pull-requests))

- [x] Tell us about the problem your pull request is solving. the current example causes the rediect to `http://login/` but it should actually just redirect to the page `/login`.
- [x] Are there any open issues that are related to this? no
- [x] Is this PR dependent on PRs in other repos? no


marshallswain:
Thank you

PR #455, opened 2020-03-29
improvement(service-module.getters): improve "get" getter

J3m5:
- Apply destructuring directly on arguments
- Only call select on "tempsById" if record is undefined and return record early otherwise
- Convert ternary to "short-circuiting"


marshallswain:
Thanks!

PR #459, opened 2020-04-10
Temp clone management

hamiltoes:
### Summary

I'm working on a simple note taking app that auto-saves after user input. I've ran into some issues with temps and clones not behaving as I expected.

  - Calling `clone.save()` multiple times creates multiple rows in the DB because clones do not get the ID after the first `create`. This can be avoided by re-cloning after `create` succeeds, but that discards any changes to the clone that occur while waiting for the API server's response. So you have to copy modified properties to a temp object, then re-clone. This is tedious and can be avoided by just adding the ID received from the `create` response to the instance's clone.
   - Similar to above, clones keep their `__isTemp` status after successful `create`. This means subsequent calls to `clone.commit()` add `__isTemp` back to the original model instance. I didn't have any major issues with this, but it made debugging more confusing.
  - I am using the `makeFindMixin` to retrieve list items for my notes. When querying with `temps: true`, I briefly get a `Duplicate keys detected ... This may cause update error.` error from Vue after `create` succeeds. This is because my `v-for`'s `key` is bound to the query result's `item.id`. But the real reason is that during the process of moving models from `tempsById` to `keyedById`, `find()` runs again and includes the model from both `tempsById` and `keyedById`. I can avoid this error by just binding the `v-for`'s `key` to the `v-for`'s index, but then I lose reactivity per item. ~The actual issue is that the `addOrUpdate` action is `await`ed when it doesn't need to be `async` at all. This wait causes the `find()` to run in the next tick which finds the model in both places and throws the duplicate keys error before `addOrUpdate` finishes moving the model.~ That only solved my issue because my API server is not sending `created` events back to the client that caused them. So I think a more general solution is to take the union (by ID or temp ID) between `keyedById` and `tempsById` in the find getter when `temps: true`.

  - no
  - no

### Other Information

n/a

### DIFF
--- a/src/service-module/make-base-model.ts
+++ b/src/service-module/make-base-model.ts
@@ -185,11 +185,7 @@ export default function makeBaseModel(options: FeathersVuexOptions) {
     }
 
     public static getFromStore(id, params?) {
-      if (params) {
-        return this._getters('get', id, params)
-      } else {
-        return this._getters('get', id)
-      }
+      return this._getters('get', id, params)
     }
 
     /**
--- a/src/service-module/service-module.actions.ts
+++ b/src/service-module/service-module.actions.ts
@@ -322,7 +322,7 @@ export default function makeServiceActions(service: Service<any>) {
         service.FeathersVuexModel &&
         !(item instanceof service.FeathersVuexModel)
       ) {
-        item = new service.FeathersVuexModel(item)
+        item = new service.FeathersVuexModel(item, { commit: false })
       }
 
       // If the item has a matching temp, update the temp and provide it as the new item.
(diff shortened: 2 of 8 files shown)


marshallswain:
Thanks @hamiltoes!  Do you consider this ready to merge?

hamiltoes:
Yep!

Although after reviewing things today, I realized the `find` getter doesn't have to change if we just move the temp directly to `keyedById` in the `updateTemp` mutation. This would actually mean we could get rid of `tempsByNewId`. Then the `addOrUpdate` action could be simplified because it doesn't need to worry about `tempsByNewId` at all. It would just add or update based on whether the item already exists in `keyedById`.

Would you prefer I submit a new PR for that or just include it here?

marshallswain:
I think I understand what you are saying. Feel free to update this PR

hamiltoes:
I think this simplifies the process of promoting a temp to a "real" item a bit

marshallswain:
Released to npm in `feathers-vuex@3.9.2`

PR #462, opened 2020-04-13
Fix: Auth module responseHandler modifying state outside of mutation

hamiltoes:
### Summary

The `responseHandler` in `auth-module.actions` first sets the entire auth response as `payload` in the auth module state, then if a user was returned and a user Model exists, a new Model is created. If the user Model's `setupInstance` changes the data (e.g. convert datetime to `Date`), then it is actually modifying `payload` state and Vuex throws an error (in `strict` mode).

### Solution

`fastCopy` the user data before creating the Model instance so that `setupInstance` runs on the copy

no

no

### DIFF
--- a/src/auth-module/auth-module.actions.ts
+++ b/src/auth-module/auth-module.actions.ts
@@ -3,6 +3,7 @@ eslint
 @typescript-eslint/explicit-function-return-type: 0,
 @typescript-eslint/no-explicit-any: 0
 */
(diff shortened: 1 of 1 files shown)


marshallswain:
Thanks for finding and fixing this issue!

marshallswain:
Released to npm in `feathers-vuex@3.9.2`

Issue #466, opened 2020-04-26
Bug: FeathersVuexFind components


Barbapapazes:
### Steps to reproduce

I copied the code from the example [server-side pagination using FeathersVuexFind components](https://vuex.feathersjs.com/data-components.html#server-side-pagination). But, when I try to change the skip value, nothing append. So I added the watch property: 
```js
watch="query"
```
In the documentation, at [distinct query and fetchQuery](https://vuex.feathersjs.com/data-components.html#use-a-distinct-query-and-fetchquery), we can read:
> the watch strings will be modified internally to watch the fetchQuery object. This means if you are watching query.userId and you add a fetchQuery, the component is smart enough to know you meant fetchQuery.userId

But, it is not exactly what happened !

I found this in the source code:
```ts
const watch = Array.isArray(this.watch) ? this.watch : [this.watch]
 if (this.fetchQuery || this.query) {
   watch.forEach(prop => {
     if (typeof prop !== 'string') {
       throw new Error(`Values in the 'watch' array must be strings.`)
     }
     if (this.fetchQuery) {
       if (prop.startsWith('query')) {
         prop.replace('query', 'fetchQuery')
       }
     }
     this.$watch(prop, this.fetchData)
   })
   this.fetchData()
 }
```

But the replace() method returns a new string ([mdn js doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace)). And the return case is not handle !

### Expected behavior
We expected that when we provided a query and a fetchQuery, the watch look at the changes of the fetchQuery, not the query.

### Actual behavior
Actually, the watch will watch the query, even a fetchQuery is provided

### System configuration

Tell us about the applicable parts of your setup.

**Module versions** (especially the part that's not working):
```
"feathers-vuex": "^3.9.1"
```

**NodeJS version**:
```
v12.14.1
```

**Operating System**:
```
Windows 10 Pro, version 1909
```

**Browser Version**:
```
Edge (Chromium), Dev channel, Version 84.0.488.1 (Official build) dev (64-bit)
```

J3m5:
Thank you!
Would you mind to submit a pull request to fix this issue?

Barbapapazes:
I'm working on it !

marshallswain:
Closing now that the PR has been merged

PR #467, opened 2020-04-26
fix issue #466 (Bug: FeathersVuexFind component)

Barbapapazes:
### Summary
issue #466 

Bug in the component [FeathersVuexFind](https://vuex.feathersjs.com/data-components.html#feathersvuexfind)

There was a problem with the watch property who was watching query, even a fetchQuery was provided. This was due to the method replace() which return a new string, and not change the string ([mdn js doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace)).

It's fix. Thank you !

### DIFF
--- a/src/FeathersVuexFind.ts
+++ b/src/FeathersVuexFind.ts
@@ -151,7 +151,7 @@ export default {
         }
         if (this.fetchQuery) {
           if (prop.startsWith('query')) {
-            prop.replace('query', 'fetchQuery')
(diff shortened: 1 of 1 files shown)


marshallswain:
Ah good catch!  Thanks!

marshallswain:
Released on npm as `feathers-vuex@3.9.2`

PR #468, opened 2020-04-26
fix(utils): GetId - Check idField first - Fix #463

### LINKED ISSUE #463: Model's idField ignored
### Steps to reproduce

This is the first time I'm using feathersjs and vuex so apologies if I've misunderstood the documentation.

I have a bikes service that interacts with my backend service. I'm able to fetch my data fine but I'm struggling to patch it back to the server.

I have the following `idField` specified on my bikes model:
```
...
const servicePath = 'bikes'
const servicePlugin = makeServicePlugin({
  Model: Bike,
  idField: 'bikeId',
  debug: true,
  service: feathersClient.service(servicePath),
  servicePath,
})
...
```
On the server the `bikeId` field is also specified: 

```
export default function (app: Application) {
  const options = {
    Model: createModel(app),
    paginate: app.get('paginate'),
    id: 'bikeId',
  };
  ...
```
After a successful `useFind` the state for bikes shows that the correct idField is set in `$store.state.bikes.idField` but that field isn't used as the id, instead it is using the `_id` field from mongo on the server:
```
{
  "ids": [
    "5e9acedc2909ac001aa7ab85",
    "5e9acedc2909ac001aa7ab87",
    "5e9acedc2909ac001aa7ab88",
    "5e9acedc2909ac001aa7ab89",
    "5e9acedc2909ac001aa7ab86",
    "5e9acedc2909ac001aa7ab8b",
    "5e9acedc2909ac001aa7ab8a",
    "5e9acedc2909ac001aa7ab8c"
  ],
  "keyedById": {
    "5e9acedc2909ac001aa7ab85": {
      "stravaAthleteId": 2164652,
      "bikeId": "b2680865",
      "order": 1,
      "_id": "5e9acedc2909ac001aa7ab85",
      "bikeName": "Canyon Dude",
      "resourceState": 2,
      "totalDistance": 3046948,
      "updatedAt": "2020-04-20T07:41:45.262Z",
      "createdAt": "2020-04-19T17:35:35.849Z"
    }
...
```

When I use the model's `save` method the id used for the patch is the long mongo id which the server ignores as it isn't using that id to match records, it is expecting the `bikeId` field's value instead. If I log the incoming requests on the server in the patch method, I can see the id is not the correct one, for example: `Incoming patch with id:  5e9acedc2909ac001aa7ab8c` 

I'm using the rest client on a Nativescript built Android app (currently just in simulator). Connections between server and client are working fine.

### Expected behavior
I would expect the model's `save` method to use the value of the configured `idField`.

### Actual behavior
Instead the model's `save` method uses the `_id` field as the id which doesn't match what the server is expecting.

### System configuration

Tell us about the applicable parts of your setup.

**Module versions**:
```
feathers-vuex@3.9.1
vuex@3.1.3
feathers@4.5.2
```

**NodeJS version**: `v12.16.1`

**Operating System**: Linux

**Browser Version**: Frontend is using Nativescript/Android

**React Native Version**: N/A. Using Vue and Nativescript

**Module Loader**: Webpack

J3m5:
Try to get the item's id with the idField first.
Fix #463

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -212,14 +212,18 @@ export function assignTempId(state, item) {
 
 /**
  * Get the id from a record in this order:
- *   1. id
- *   2. _id
- *   3. the `idField`
+ *   1. the `idField`
+ *   2. id
+ *   3. _id
  * @param item
+ * @param idField
  */
 export function getId(item, idField) {
   if (!item) {
     return
+  }
+  if (item[idField] != null || item.hasOwnProperty(idField)) {
+    return item[idField]
   }
   if (item.id != null || item.hasOwnProperty('id')) {
     return item.id
@@ -227,9 +231,7 @@ export function getId(item, idField) {
   if (item._id != null || item.hasOwnProperty('_id')) {
     return item._id
   }
-  if (item[idField] != null || item.hasOwnProperty(idField)) {
-    return item[idField]
-  }
+
 }
 
 // Creates a Model class name from the last part of the servicePath


marshallswain:
Thanks!

marshallswain:
Released to npm as `feathers-vuex@3.9.2`

Issue #472, opened 2020-04-30
Problem with handleEvents option in makeServicePlugin method


s3ns3i:
I have a problem, that is described there: https://github.com/feathersjs-ecosystem/feathers-vuex/issues/397
so I thought, that for now I would refresh other services by using _handleEvents_ property, but it seems to not appear in the state of the model (I don't know if it is supposed to), but also it's not working (I put console.log there, and it's not called).

### Steps to reproduce

This is what I did:

```
// store/services/roles.js
const servicePlugin = makeServicePlugin({
  Model: Role,
  service: feathersClient.service(servicePath),
  servicePath,
  handleEvents: {
    created: (item, { model, models }) => {
      console.log("roles: created event");
      return true;
    },
    ... // Same with the rest of handlers
  }
});
```

When it wasn't working, I also added this part:

```
// plugins/feathers.js
const {
  makeServicePlugin,
  makeAuthPlugin,
  BaseModel,
  models,
  FeathersVuex
} = feathersVuex(feathersClient, {
  serverAlias: "api"
  idField: "id" // Using sequelize
  debug: true,
  whitelist: ["$regex", "$options"],
  enableEvents: true,
  handleEvents: {
    created: () => true,
    updated: () => true,
    patched: () => true,
    removed: () => true
  }
});
```

Is there something I am doing wrong? It seems straightforward, but no luck so far.
If you will need more details, I will try to create a bigger example

### Expected behavior
Handlers are called when I create an object.

### Actual behavior
Handlers are not called.

### System configuration

**Module versions**

    "feathers-vuex": "^3.9.4",
    "@feathersjs/feathers": "^4.5.3",
    "@feathersjs/socketio-client": "^4.5.4",
    "@vue/composition-api": "^0.5.0",
    "feathers-hooks-common": "^5.0.3",
    "nuxt": "^2.12.2",
    "socket.io-client": "^2.3.0",

**NodeJS version**: v13.13.0

**Operating System**: Ubuntu 19.10

**Browser Version**: Mozilla 75.0

J3m5:
Can you please check in your network panel if you are receiving the events ?
I don't see how it couldn't works...

s3ns3i:
I don't see anything in my network panel, but when I create a role on one browser, it shows on the other one, so I assume, that sockets work.

I've also added a code from this section: https://vuex.feathersjs.com/model-classes.html#model-events - this one works (I can see logs in the console).

[...]

PR #473, opened 2020-04-30
fix: don't overwrite handleEvents handlers #472

hamiltoes:
### Summary

`handleEvents` handlers were being overwritten. See #472 .

  -  #472 
  - no

### Other Information

I also added falling back to any globally configured `handleEvents` handlers if configured. This will allow services to have specific `handleEvents` handlers for only some instead of all events and then fall back to an app-wide implementation (if one is provided) for the other events.


marshallswain:
Beautiful. Thanks!

PR #476, opened 2020-05-02
feature: handle complex events

hamiltoes:
### Summary

In my application, I would like to pass extra information with service events from the API server (e.g. who caused the event, what device they're on, etc.) so that when data changes, a component can indicate to the user that the content was modified by user Foo or by device "Bar's IPhone". Currently, this is not really workable as the entire event data is passed to the `addOrUpdate` or `removeItem` actions. I could modify the event data directly in a `handleEvents` handler, but this would also change what data is passed to `Model.emit()` and therefore defeat the purpose of passing event source/context with the event to components listening with `Model.on()`.

My proposal is to allow the `handleEvents` handler to somehow modify what data is dispatched to actions, but still pass the entire event data to `Model.emit()`.

I originally thought about just returning the object to dispatch but did not want to deal with discerning what values for `affectsStore` are truthy and what values should actually be dispatched. Instead, my implementation allows returning a tuple `[affectsStore, dataToDispatch]` where `affectsStore` behaves exactly as it does now (can be truthy), and `dataToDispatch` is the data to dispatch to actions. To preserve current functionality and not make this a breaking change, if an array is not returned from the handler, the value is used for `affectsStore` and the original item is dispatched.

### Other Information

  - no
  - this PR depends and is based on #473 in this repo
  - if accepted, this would need an update in the docs repo. I will submit a PR there once I receive feedback on this PR.

### DIFF
--- a/src/service-module/make-service-plugin.ts
+++ b/src/service-module/make-service-plugin.ts
@@ -64,7 +64,7 @@ export default function prepareMakeServicePlugin(
     } = options
 
     events.forEach(eventName => {
-      if (options.handleEvents[eventName])
+      if (!options.handleEvents[eventName])
         options.handleEvents[eventName] = () => options.enableEvents || true
     })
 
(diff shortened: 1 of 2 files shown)


marshallswain:
Lovely and elegant solution.  Thank you for this!

marshallswain:
Is event handling even documented in any place?

marshallswain:
@hamiltoes, could you open another PR with some documentation?  If not a PR, then please an issue so I don't drop this like the auth tests.  ;). Thanks!

PR #479, opened 2020-05-05
fix(FeathersVuexFind.ts): Fixes the use of the lodash "get" method

### LINKED ISSUE #364: FeathersVuexFind - server-side pagination stops returning pageInfo.ids when $in operator is used in query
### Steps to reproduce

Setup a basic FeathersVuexFind component using server-side pagination - [example in docs](https://vuex.feathersjs.com/data-components.html#usage-examples).

Modify the query so at least one property uses the $in operator. In this example, the API is setup to have a bunch of todos with the title "test" - so this does return results:

```
<template>
  <FeathersVuexFind
    :service="service"
    :query="internalQuery"
    :fetch-query="fetchQuery"
    :edit-scope="getPaginationInfo"
  >
    <div slot-scope="{ items: todos }">
      {{todos}}
    </div>
  </FeathersVuexFind>
</template>

<script>
  export default {
    data() {
      return {
        service: 'todos',
        ids: [],
        query: {
          title: {
            $in: ['test'] // this breaks things?
          }
        },
        total: 0,
        limit: 10,
        skip: 0
      };
    },
    computed: {
      internalQuery() {
        const { idField } = this.$store.state[this.service];
        return {
          [idField]: {
            $in: this.ids
          }
        };
      },
      fetchQuery() {
        return Object.assign({}, this.query, { $limit: this.limit, $skip: this.skip });
      }
    },
    methods: {
      getPaginationInfo (scope) {
        const { queryInfo, pageInfo } = scope;
        console.log(scope)
        this.total = queryInfo.total;
        if (pageInfo && pageInfo.ids) {
          console.log(pageInfo.ids)
          this.ids = pageInfo.ids;
        } else {
          console.warn('pageInfo is missing ids')
        }
      }
    }
  }
</script>
```

### Expected behavior
Server returns the results. The ids are available within the `getPaginationInfo` function.

### Actual behavior
Server returns the results, but within `getPaginationInfo`, `scope.pageInfo` is an empty object. `scope.items` is also an empty array.

### System configuration
Client:
```
"feathers-client": "^2.4.0",
"feathers-vuex": "^2.2.4"
```
Server:
```
"@feathersjs/feathers": "^4.3.7",
"feathers-objection": "^4.4.3",
"knex": "^0.20.1",
"objection": "^1.6.11",
```
### Notes

- Data is returned from the API (and present in the `updatePaginationForQuery` mutation). 
- The same query when run against `Post.findInStore(query)` and `Post.find(query)` returns the results OK.
- Without the $in operator, it works fine.

Am I missing something obvious here or is this a bug?

### LINKED ISSUE #423: empty pageInfo in query using $and or $or
### Steps to reproduce

i'm following the renderless data component docs example using normal query like:
`{ $limit, $skip, [fieldname]: { $search: string } }`
work just fine.

but, if i try
`{ $limit, $skip, $or: [ { [fieldname]: { $search: string } } ] }`
then my client get the data but pageInfo is empty

i don't know if i miss something or there is something wrong in my code & setup

### Expected behavior
working just fine

### Actual behavior
no pageInfo

### System configuration

**Client Package**
```
    "@feathersjs/feathers": "^4.5.1",
    "feathers-hooks-common": "^5.0.2",
    "feathers-vuex": "^3.5.5",
```

**Server Package**
```
    "@feathersjs/express": "^4.5.1",
    "@feathersjs/feathers": "^4.5.1",
```

J3m5:
Should fix #364 and fix #423 but it needs testing.

### DIFF
--- a/src/FeathersVuexFind.ts
+++ b/src/FeathersVuexFind.ts
@@ -67,7 +67,7 @@ export default {
     },
     queryInfo() {
       if (this.pagination == null || this.queryId == null) return {}
-      return _get(this.pagination, `[${this.queryId}]`) || {}
+      return _get(this.pagination, this.queryId, {})
     },
     pageInfo() {
       if (
@@ -76,7 +76,7 @@ export default {
         this.pageId == null
       )
         return {}
-      return _get(this.pagination, `[${this.queryId}][${this.pageId}]`) || {}
+      return _get(this.pagination, [this.queryId, this.pageId], {})
     },
     scope() {
       const { items, isFindPending, pagination, queryInfo, pageInfo } = this


J3m5:
@marshallswain I made a quick search in the repo and it seems that you are using this bracket syntax with the lodash get method at other places, is there a specific reason to use it ?

Unless the key you want to access is a string wrapped in brackets, I don't see any.

If we look at [the lodash documentation](https://lodash.com/docs/4.17.15#get), the brackets are only used to access array indexes and it's not mandatory, you can simply do 
```js
_.get(object, 'a.0.b.c'); 
// or
_.get(object, ["a", 0, "b", "c"]);

```

marshallswain:
Let's get them all fixed!  Not sure how much sleep I had lost to mess this up, but the simpler the better.  Thanks for the PR

J3m5:
You're welcome!
I'll submit an another PR to fix them all!

PR #481, opened 2020-05-07
useFind returned attributes updated

amoghpalnitkar:
### Summary

Updated the docs at `docs/composition-api.md.`

The Interface descriptor shows the latest return values for useFind method. However the description for each return value still had the older value names. Updated the ones in the description to match the latest names.

There isn't any open issue for this PR.


marshallswain:
Thanks!

PR #483, opened 2020-05-12
Adding new model functions count() and countInStore()

tylermmorton:
### Summary

(If you have not already please refer to the contributing guideline as [described
here](https://github.com/feathersjs/feathers/blob/master/.github/contributing.md#pull-requests))

Hi all, forgive me in advance as this is my first PR. I wanted to try and make a contribution to Feathers-Vuex since I've been using it with Nuxt so much lately! Hopefully this won't be my last 👍 

- [ ] Tell us about the problem your pull request is solving.

This addition makes it easier to get a total count of the records stored on the Feathers server and a total count of the records stored by the Vuex store. 

My use case was needing the total count of posts on my CMS dashboard. Instead of using the fast-count method `Post.find({query: {$limit: 0}})` I added a quick and easy utility for it to make it cleaner and reduce nesting. I've also updated the unit tests and the documentation. 

No

No

### DIFF
--- a/docs/model-classes.md
+++ b/docs/model-classes.md
@@ -5,17 +5,15 @@ sidebarDepth: 3
 
 # Data Modeling with Model Classes
 
-Feathers-Vuex 1.0 introduced some lightweight data modeling. Every service had its own, internal `FeathersVuexModel`. In version 2.0 this `FeathersVuexModel` is now called the `BaseModel` and is extendable, so you can add your own functionality.
+Feathers-Vuex 1.0 introduced some lightweight data modeling.  Every service had its own, internal `FeathersVuexModel`.  In version 2.0 this `FeathersVuexModel` is now called the `BaseModel` and is extendable, so you can add your own functionality.
+
 
 ## Extending the BaseModel Class
 
-While [setting up Feathers-Vuex](/getting-started.html#feathers-client-feathers-vuex), we exported the `BaseModel` class so that we could extend it. The below example shows how to import and extend the `BaseModel`. Each service must now have its own unique Model class.
+While [setting up Feathers-Vuex](/getting-started.html#feathers-client-feathers-vuex), we exported the `BaseModel` class so that we could extend it.  The below example shows how to import and extend the `BaseModel`.  Each service must now have its own unique Model class.
 
 ```js
-import feathersClient, {
-  makeServicePlugin,
-  BaseModel
-} from '../feathers-client'
+import feathersClient, { makeServicePlugin, BaseModel } from '../feathers-client'
 
 class User extends BaseModel {
   // Required for $FeathersVuex plugin to work after production transpile.
@@ -37,7 +35,7 @@ const servicePlugin = makeServicePlugin({
 })
 ```
 
-In case you're wondering, the `modelName` property is used to get around transpilation errors when using Babel with ES3 or ES5. Babel is still installed by default in most projects and generators. The `modelName` is used instead of the `name` property to provide a reliable name AFTER transpilation.
+In case you're wondering, the `modelName` property is used to get around transpilation errors when using Babel with ES3 or ES5.  Babel is still installed by default in most projects and generators.  The `modelName` is used instead of the `name` property to provide a reliable name AFTER transpilation.
 
 If you're working in an environment that doesn't support static properties on classes, you can always specify the static properties using the dot operator:
 
@@ -45,7 +43,7 @@ If you're working in an environment that doesn't support static properties on cl
 class User extends BaseModel {}
 
 User.modelName = 'User'
-User.instanceDefaults = function () {
+User.instanceDefaults = function() {
   return {
     email: '',
     password: ''
@@ -77,7 +75,7 @@ created () {
 
 ### findInStore(params)
 
-Model classes have a `findInStore` method, which is a proxy to the [`find` getter](./service-plugin.html#Service-Getters). <Badge text="1.7.0+" />
+Model classes have a `findInStore` method, which is a proxy to the [`find` getter](./service-plugin.html#Service-Getters).  <Badge text="1.7.0+" />
 
 ```js
 // In your Vue component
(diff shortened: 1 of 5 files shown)


fratzinger:
Beautiful first PR! Already with tests and docs. (I'm also just a guy who uses FeathersVuex a lot)

I also thought of this feature before, but would take it a step further. To count only all available items is a limited use case for me. An optional `params` argument would fit my needs. It would be the same like the `find`-getter or `find`-action with a default `params = { query: {} }` and `$limit: 0` used with `Object.assign(defaultParams, params, { query: { $limit: 0 } })`.
In my head, my solution could be used the exact same way you implemented it.

I think it's also a good use case for a renderless component as a badge for e.g. notifications, unread messages. That would be really neat.

What do you think? Do you guys think it's a good addition? If so, I would be happy to implement it or help with it.

J3m5:
Indeed, this is a great PR @tylermmorton !

I was thinking about that too @fratzinger. But maybe we could just pass the query and not all the params, otherwise it would save us only 9 characters.
That would mean we could call it like that:
```js
Post.count({read: false})
```
The downside is that we can't pass custom params to the server.
What do you think ?

J3m5:
For the countInStore getter, you could call the find getter only if params are present.

fratzinger:
> `Post.count({read: false})`
> The downside is that we can't pass custom params to the server.
Yes, the 9 characters are not really worth the effort, you're right. The other more important downside (IMO) with your suggestion is that it leads to misunderstandings when to pass `params` and when to pass `query` to a FeathersVuex-methods.
With a renderless component we could make a `:query=""`-prop which makes that clear.
On the other hand the 9 characters (`$limit: 0`) is not as inuitive as a `count`-method. I, for myself, had to google the solution and found the small passage in the feathers-docs. I was close to just make an ordinary request and use `.length` or `.total`.

> For the countInStore getter, you could call the find getter only if params are present.
Indeed. I'd suggest: 
```
count: (state, getters) => (params) => {
  params = params || { query: {} };
  params.$limit = 0;
  return getters.find(params).data.length;
},
```

J3m5:
This misunderstanding has cost me many errors because I often pass directly the query and forget the to wrap in a params property.
But, after thinking about it, we need the params in the find getter if we want [to count the temps records too](https://github.com/feathersjs-ecosystem/feathers-vuex/blob/0af97a1caf4f7e50f130e2bbeddbdcc3f2de84fd/src/service-module/service-module.getters.ts#L45).

```js 
if (params.temps) {
  values = values.concat(_.values(state.tempsById))
}
```

If we take this into account, we end up with this:

```js
Post.count({ query:{ read: false }})
```
An other solution would be to support both case by checking if the query prop is here or not, but it wold be inconsistent with the rest of the codebase.

I wouldn't be against applying this logic everywhere but this is out of the scope of this PR.

src/service-module/make-base-model.ts
```diff
@@ -176,6 +176,19 @@ export default function makeBaseModel(options: FeathersVuexOptions) {
       return this._getters('find', params)
     }
 
+    public static count() {
+      const params = {
+        $limit: 0 // <- limit 0 in feathers is a fast count query
```
hamiltoes, line 181: Tiny correction: `$limit` needs to be wrapped in the `query`
```suggestion
        query: { $limit: 0 } // <- limit 0 in feathers is a fast count query
```

tylermmorton: Good catch! Thanks

tylermmorton:
Hey guys, thanks for all of the great feedback. I definitely like the idea of allowing the user to pass params to the `count()` function if they want. At that point it does become an alias for a call to `find()` with the `$limit: 0` query param, but to me it becomes a lot more semantically clear when `count()` is used.

src/service-module/make-base-model.ts
```diff
@@ -176,6 +176,20 @@ export default function makeBaseModel(options: FeathersVuexOptions) {
       return this._getters('find', params)
     }
 
+    public static count(params) {
+      params = params || {
+        query: {} 
+      }
```
fratzinger, line 182: `if (!params.query) { throw "params must contain a query-object" }`
for when `params` is provided, but with no `query`

src/service-module/make-base-model.ts
```diff
@@ -176,6 +176,20 @@ export default function makeBaseModel(options: FeathersVuexOptions) {
...
+      params.query.$limit = 0; // <- limit 0 in feathers is a fast count query
+      return this._dispatch('find', params).then((res) => {
+        return res.total
```
fratzinger, line 185: `return res.total || res.length`
if server has no pagination ;)

src/service-module/service-module.getters.ts
```diff
@@ -21,6 +21,35 @@ export default function makeServiceGetters() {
     list(state) {
       return state.ids.map(id => state.keyedById[id])
     },
+    count: state => params => {
```
fratzinger, line 24: why not just?:
```
count: (state, getters) => params => {
  if (isRef(params)) {
    params = params.value
  }
  params = params || { query: {} }
  return getters.find(params).data.length;
}
```

docs/model-classes.md
```diff
@@ -85,6 +85,32 @@ created () {
 }
 ```
 
+### count(params)
+
+Model classes have a `count` method, which is a proxy to the [`find` action] with the param `$limit: 0` automaticalluy appended. On the Feathers server, `$limit: 0` results in a fast count query. (./service-plugin.html#find-params). <Badge text="3.10.5+" />
```
marshallswain, line 90: Small typo here: automaticalluy --> automatically

src/service-module/make-base-model.ts
marshallswain: I like the idea of an explicitly-required `query: {}`.  There's no convenience in discovering that you've accidentally queried everything with the empty set.

marshallswain:
Lol.  I didn't even know that `{ $limit: 0 }` would perform a count, so I've never used it.  This is a great PR.  Please @tylermmorton, let me know when you feel like it's ready for a merge.

marshallswain:
Hey, would one of you guys be available to review this and see if it feels production ready.  If none of you have the bandwidth right now, no worries.  I'll get to it in a couple of weeks at the earliest.  Thanks!

@fratzinger @hamiltoes @J3m5 @tylermmorton

fratzinger:
I'm on it. 

I'm changing and adding a few things. I'll create a seperate PR with a count action and a FeathersVuexCount renderless component within the next days.

fratzinger:
see #521

PR #485, opened 2020-05-13
fix: diffOnPatch

fratzinger:
Since v3.9.0 the diffOnPatch is not fired when `params` is not defined. This little change should fix this.


marshallswain:
Heh.  Good catch.  I was looking at this line, yesterday, thinking something was off with it.  Thanks for the fix!

PR #487, opened 2020-05-14
typescript: add base model typing

hamiltoes:
### Summary

Add explicit typing to BaseModel to aide development.

  Currently, the BaseModel has type `any` which makes intellisense nearly useless. I am adding interfaces with jsdoc strings to further improve the already great experience of developing with FeathersVuex.
  no
  no

### Other Information

Because the BaseModel class is constructed inside a function, it's type cannot be used outside of that function. My solution is to define multiple interfaces: one for the static side `ModelStatic`, and one for the instance side `ModelInstance`. Both of these interfaces are generic and take a type parameter that describes the service's data structure. There is also a `ModelInstanceClone` interface for clones.

Ideally, we would just make the BaseModel returned from the FeathersVuex setup function generic, but I could not find a way to cast the non-generic BaseModel of type `any` to a generic interface. So my solution is to return BaseModel casted to a `ModelStatic<{}>` and an additional method `castBaseModel` from the setup function which can be used to add typing for each service's data interface.

#### Notes

* `Model<D>` is of type `ModelInstance<D> & Readonly<D>`
  * This enforces clone & commit pattern because the model data is readonly so TS will error when trying to directly alter model data.
* `ModelClone<D>` is of type `ModelInstanceClone<D> & D`
  * I hide `commit()` and `reset()` methods in `ModelInstanceClone` since they just error on non-clones.
  * Clone model data is not readonly so you can directly assign to it

### Examples

If you want to check out how this works with intellisense, take a look at the following. Or even better, clone this branch, paste it in `index.ts` and play with it :-)

```ts
const { BaseModel, castBaseModel } = feathersVuex({}, { serverAlias: 'api' })

// Old way of doing things
class OldFashionedNote extends BaseModel {
  static readonly servicePath: 'old-fashioned-note'
}
const ofn = new OldFashionedNote()

// New way with types
interface NoteRecord {
  id: number
  text: string
  userId: number
  color: string
  isItBetter?: true
}

class FancyTypedNote extends castBaseModel<NoteRecord>() {
  static readonly servicePath = 'fancy-typed-note'

  static setupInstance(data: Partial<NoteRecord>, ctx: SetupContext) {
    const typedStoreAccess = ctx.store['fancy-typed-note'].isFindPending
    return data
  }
}
const ftn = new FancyTypedNote()
ftn.text = 'TS shows error because data is readonly!'

const ftnClone = ftn.clone()
ftnClone.text = 'Direct assignment to clones is fine!'
```

#### Screenshots

Autocomplete for `OldFashionedNote` instance
* No `commit()` or `reset()`
* also no typing for actual model
oldschoolnoteinstance

Autocomplete for `OldFashionedNote` class
* includes EventEmitter methods for Model events
oldschoolnotestatic

Autocomplete for `FancyTypedNote` instance
* includes `NoteRecord` fields!
fancytypednoteinstance

### Augmentable interfaces

Users can also augment interfaces defined for `feathers-vuex` to add more helpful typing

```ts
interface MyApiModels {
  FancyTypedNote: typeof FancyTypedNote
  OldFashionedNote: typeof OldFashionedNote
}

// This example is pasted in `index.ts` but a user would actually use
// declare module 'feathers-vuex' {...
declare module '.' {
  interface FeathersVuexStoreState {
    'old-fashioned-note': ServiceState
    'fancy-typed-note': ServiceState
  }
  interface FeathersVuexGlobalModels {
    api: MyApiModels
  }
}
```
And then they can get autocomplete on `Model.models`, `Model.store`, `Vue.$FeathersVuex`, setupInstance's `store` and `models` params, etc.

### DIFF
--- a/docs/model-classes.md
+++ b/docs/model-classes.md
@@ -51,9 +51,9 @@ User.instanceDefaults = function() {
 }
 ```
 
-### BaseModel typing <Badge text="3.11.0+" />
+### BaseModel typing
 
-Version `3.11.0` brings explicit typing to the BaseModel. This gives helpful IDE autocomplete and errors when using Model classes.
+BaseModel typing gives helpful IDE autocomplete and errors when using Model classes.
 
 Since Feathers-Vuex doesn't know what your data looks like, you will need to help define your underlying model data's interface.
 
--- a/docs/model-classes.md
+++ b/docs/model-classes.md
@@ -15,7 +15,7 @@ While [setting up Feathers-Vuex](/getting-started.html#feathers-client-feathers-
 ```js
 import feathersClient, { makeServicePlugin, BaseModel } from '../feathers-client'
 
-class User extends BaseModel {
+export class User extends BaseModel {
   // Required for $FeathersVuex plugin to work after production transpile.
   static modelName = 'User'
   // Define default properties here
@@ -53,85 +53,74 @@ User.instanceDefaults = function() {
 
 ### BaseModel typing <Badge text="3.11.0+" />
 
-Version `3.11.0` brings explicit typing to the BaseModel. This gives helpful IDE autocomplete and errors when Model classes or instances
-are used incorrectly.
+Version `3.11.0` brings explicit typing to the BaseModel. This gives helpful IDE autocomplete and errors when using Model classes.
 
-One major change here is that the BaseModel now enforces Vuex strict mode compliance by default and encourages proper state management using
-the clone and commit pattern. It does this by declaring the underlying model data `readonly`. This means the TS compiler will error at any
-attempt to directly assign to a model.
+Since Feathers-Vuex doesn't know what your data looks like, you will need to help define your underlying model data's interface.
 
-Take the `User` class from above
+By default, Model classes are `string` indexable with value of type `any`. This isn't super helpful...
 
 ```ts
-// wrong
-const user = new User()
-user.email = 'harry.potter@hogwarts.edu' // <- TS will error here
+// Just like before, we define our class as an extension of BaseModel
+export class User extends BaseModel { /* ... */ }
+
+// Augment the User Model interface
+export interface User {
+  email: string
+  password: string
+}
 ```
 
-The proper way to edit an existing `User` instance is to clone it, edit the *clone's* props, and then commit the changes.
-This ensures changes don't propagate to the rest of the app until ready.
+Now, whenever we access a `User` model, all fields defined in the interface will be available in IDE auto-complete/intellisense along with the model methods/props.
+
+If you already have a User interface defined under a different name, just define a new interface with the same name as your Model class like so
 
 ```ts
-// correct
-const clone = user.clone()
-clone.email = 'harry.potter@hogwarts.edu' // <- No error here
-clone.commit()
+// if our User interface already exists as UserRecord
+export interface User extends UserRecord {}
 ```
 
-You can disable this `readonly` behavior if desired. In `feathers-client.ts`, augment FeathersVuex's typing
+To further enhance typing, you can augment FeathersVuex types to aid development in other parts of your app. It's important to note the differences in the following example if you do or do not setup a `serverAlias`.
 
 ```ts
-declare module 'feathers-vuex' {
-  interface FeathersVuexTypeOptions {
-    'model-readonly': false
+// src/store/user.store.ts
+import { ServiceState } from 'feathers-vuex'
+
+export class User extends BaseModel { /* ... */ }
+export interface User { /* ... */ }
+const servicePath = 'users'
+
+declare module "feathers-vuex" {
+  interface FeathersVuexStoreState {
+    [servicePath]: ServiceState<User>
+  }
+
+  // Only if you setup FeathersVuex without a serverAlias!!
+  interface FeathersVuexGlobalModels {
+    User: typeof User
+  }
+}
+
+// Only if you setup FeathersVuex with a serverAlias!!
+declare module "src/store" {
+  interface MyApiModels {
+    User: typeof User
   }
 }
 ```
 
-### Casting the BaseModel <Badge text="3.11.0+" />
-
-Typescript users can further enhance typing on Model classes and instances by passing their data's underlying structure as an interface
-to `castBaseModel<T>()`. This gives helpful type hints and autocomplete from your IDE when interacting with your underlying Model data.
-
-To take advantage of this, first we need to update `feathers-client.ts` to export the new function
+If you have setup a `serverAlias`, you need to add the following to `src/store/index.ts`.
 
 ```ts
-// feathers-client.ts
-const {/* other props, */ castBaseModel } = feathersVuex(/* ... */)
-
-// Export `castBaseModel` too
-export { /* other props */ castBaseModel }
+// src/store/index.ts
+export interface MyApiModels { /* Let each service augment this interface */ }
+declare module "feathers-vuex" {
+  interface FeathersVuexGlobalModels {
+    'my-api-name': MyApiModels
+  }
+}
 ```
 
-Now we can use `castBaseModel()` when defining our Model class
-
-```ts
-import feathersClient, { makeServicePlugin, castBaseModel } from '../feathers-client'
-
-// Define an interface for your underlying data
-interface UserData {
-  id: number
-  email: string
-  password: string
-}
-
-// Pass interface to castBaseModel()
-class TypedUser extends castBaseModel<UserData>() {
-  static readonly modelName = 'TypedUser'
-}
-
-// Make the plugin just as before
-const servicePath = 'users'
-const servicePlugin = makeServicePlugin({
-  Model: TypedUser,
-  service: feathersClient.service(servicePath),
-  servicePath
-})
-```
-
-Now our IDE knows that instances of `TypedUser` have all props defined by the `UserData` interface.
-
-By default, the BaseModel uses `{ [key: string]: any }` as the underlying data interface meaning any prop can be accessed/assigned to.
+Replace `my-api-name` with the `serverAlias` you used when setting up FeathersVuex.
 
 ## Model attributes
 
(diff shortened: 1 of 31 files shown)


hamiltoes:
I have intentionally left out some methods and fields from the BaseModel class but if I am mistaken and those should be publicly visible, please let me know! (Or if other should be hidden)
* `hydrateAll()` - I've never used SSR but wasn't sure if this was called internally or by clients
* `keepCopiesInStore` - this is going to be deprecated
* `merge()` - under the impression this is internal use only

I'm probably forgetting a few...

J3m5:
This is awesome! Thank you for your contribution.

It seems that there is a extra brace at the last line of the index.ts file and it makes the CI build fail.

hamiltoes:
@J3m5 ah thank you! That's what I get for making my playground in `index.ts` 😛

J3m5:
There are a lot of Typescript errors that came up [in the CI build](https://travis-ci.org/github/feathersjs-ecosystem/feathers-vuex/builds/687335837#L273), can you take a look at it ?
Can they be caused by by the changes in the tsconfig.json file ?

hamiltoes:
@J3m5 yeah they definitely are. I think stricter null checking is good but enabling that is outside the scope of this PR so I'll revert that.

hamiltoes:
So the next issue I want some opinions on is whether or not Model data should be readonly. This will a decent amount of test updates which I'm willing to do, but don't want to do if we decide to not make it readonly :-)

Pros of making it readonly:
- enforces clone and commit pattern during development
- enforces Vuex strict mode compliance

Cons:
- will be a breaking change for everyone not using clone & commit
- adhering to vuex strict mode remains in the user's cognitive load instead of TS reminding them "hey you shouldn't directly edit this without cloning first"

Thoughts?

J3m5:
Will this only impact typescript users ?
And could this be manually changed by overriding the types ?
( I'm not very familiar with Typescript. )

hamiltoes:
Correct, only TS users will be affected by this. You can always cast away the readonly if you want
```ts
const n = new Note()
n.text = 'this throws an error' // <-- TS error: NoteRecord props are readonly

const casted = n as ModelClone<NoteRecord>
casted.text = 'this is fine' // no error
```

I will have to think about whether we could make this something the user can augment so they could globally disable it if they don't want it and don't want to cast everywhere.

J3m5:
If this can be globally disabled, I think we can enforce it by default and add some info in the docs.
And maybe add those info as JSDocs directly, no ?

hamiltoes:
There is now an exported interface `FeathersVuexTypeOptions` which users can augment. Specifically, users can now do the following to disable readonly model data

```ts
declare module 'feathers-vuex' {
  interface FeathersVuexTypeOptions {
    'model-readonly': false
  }
}
```

if `model-readonly` is anything other than false, model data will be readonly.

J3m5:
It looks good !

It seems that some tests like [this one](https://github.com/hamiltoes/feathers-vuex/blob/73f2a9fae82c4f7c6ad2620e9de15b2ea4387ff5/test/service-module/model-base.test.ts#L133) needs to be changed in order to follow the readonly constraint of the modelName on the BaseModel.

And [here](https://github.com/hamiltoes/feathers-vuex/blob/73f2a9fae82c4f7c6ad2620e9de15b2ea4387ff5/test/service-module/model-base.test.ts#L224) the FeathersVuexGlobalModels need to be augmented, [like mentionned here](https://github.com/hamiltoes/feathers-vuex/blob/73f2a9fae82c4f7c6ad2620e9de15b2ea4387ff5/src/service-module/types.ts#L74), if I understand correctly.

And I've noticed that the globalModels object is casted as FeathersVuexGlobalModels in some place but not in others, is this intended ?

hamiltoes:
> And here the FeathersVuexGlobalModels need to be augmented, like mentionned here, if I understand correctly.

I am just casting models `any` in the tests because you can't augment for each test individually. And the classes in each test are private (local to that test) so they can't be augmented to the shared global interface.

> And I've noticed that the globalModels object is casted as FeathersVuexGlobalModels in some place but not in others, is this intended ?

My goal is to make the user-facing models interface show what models they have defined (and augmented into the global interface). However, if I make the internal models of the same type, then when classes extend from BaseModel, they end up circularly referencing themselves through the `FeathersVuexGlobalModels` interface.

hamiltoes:
6c73939 makes it so that
- if a user doesn't augment `FeathersVuexStoreState`, the store has type `any`
- if a user doesn't augment `FeathersVuexGlobalModels`, the models has type `any`

This should make it easier to transition (less breaking changes) since those values have type `any` currently. It makes explicit typing of store and global models opt-in instead of required.

marshallswain:
Damn, guys!  Sweet update.  Let me know when it's ready.  It looks like it will really improve DX.  <3

marshallswain:
@hamiltoes will it improve DX if I merge this as is?

hamiltoes:
@marshallswain no, sorry. I haven't found time to finish up this PR.

I've ran into some issues that make this not work when you have a custom base class. You can't use a class type argument in a base class expression.

```ts
class MyBaseClass<T> extends castBaseModel<T>() {} // <--TS Error
```

So unfortunately I think this will be a case of over-promise, under-deliver. :-(

My plan is to pare this down to just non-generic `Model` and `ModelStatic` interfaces so we at least get typing for all the built-in methods and props.

docs/model-classes.md
```diff
@@ -51,6 +51,77 @@ User.instanceDefaults = function() {
 }
 ```
 
+### BaseModel typing <Badge text="3.11.0+" />
```
hamiltoes, line 54: Not sure if this version number will be accurate. Maybe better to just do away with the badge?

src/service-module/make-base-model.ts
```diff
@@ -38,48 +47,51 @@ export default function makeBaseModel(options: FeathersVuexOptions) {
...
-    public static copiesById = {}
+    public static readonly models = globalModels as GlobalModels // Can access other Models here
+    public static copiesById: {
+      [key: string]: Model | undefined
+      [key: number]: Model | undefined
+    } = {}
 
     public __id: string
     public __isClone: boolean
-    public data: Record<string, any>
```
hamiltoes, line 77: As far as I can tell, data was not being used anywhere so I removed it. I thought perhaps it was a remnant from a previous implementation? Can add back if necessary.

src/service-module/types.ts
```diff
@@ -59,3 +64,268 @@ export interface MakeServicePluginOptions {
...
+   *
+   * Default: `true`
+   */
+  merge?: boolean
+}
+
+export type AnyData = { [key: string]: any }
+
+/** Static Model interface */
+export interface ModelStatic extends EventEmitter {
```
hamiltoes, line 120: I've marked static props `readonly` but noticed while adding documentation that [adding static props to the class via the dot operator](https://vuex.feathersjs.com/model-classes.html#extending-the-basemodel-class) is documented and therefore probably should stay supported.

Should I make these props mutable or force clients to override with `//@ts-ignore`?

src/service-module/types.ts
```diff
@@ -59,3 +64,268 @@ export interface MakeServicePluginOptions {
...
+   * A proxy for the `get` getter
+   * @param id ID of record to retrieve
+   * @param params Get params
+   */
+  getFromStore<M extends Model = Model>(id: Id, params?: Params): M | undefined
+}
+
+/** Model instance interface */
+export interface Model {
+  [key: string]: any
```
hamiltoes, line 261: This allows indexing Models by any string and the value will have type `any`.

Ideally I would get rid of this line and force people to define an interface for the model's underlying data (as I outlined in the new docs), but that would make this a breaking change so probably best to keep this and make stricter-typing opt-in instead of breaking+opt-out.

src/useGet.ts
```diff
@@ -61,32 +61,31 @@ export default function get(options: UseGetOptions): UseGetData {
...
-    isLocal: local
-  })
+    servicePath: computed(() => model.servicePath)
+  }
+
+
 
-  function get<T>(id, params?: Params): T {
-    const idToUse = isRef(id) ? id.value : id
+  function get(id: Id, params?: Params): Promise<M | undefined> {
```
hamiltoes, line 87: One of the tests in `test/use/get.test.ts` called `get()` with no params. Should `id` be an optional parameter and `idToUse` fall back to the `id` passed in to `useGet()`?

hamiltoes: Resolving for now. This is outside the scope of this PR.

hamiltoes:
This is ready to merge pending feedback on `readonly` static props. Any thoughts @marshallswain, @J3m5 ?

J3m5:
I think we should let them as they are now, at least for now. What do you think @marshallswain ?

src/service-module/types.ts
marshallswain: @hamiltoes Let's leave them mutable for now.

marshallswain:
If static props are still mutable, then I'll go ahead and merge this.

hamiltoes:
@marshallswain I've made all static props mutable except for `store`, `models`, and `copiesById`.

hamiltoes:
This would address #519

hamiltoes:
Brought up to date with latest master. I believe this is ready to merge if there is still interest.

src/useGet.ts
```diff
@@ -61,32 +61,31 @@ export default function get(options: UseGetOptions): UseGetData {
...
         ? params
         : { ...params }
-      if (getterParams != null) {
-        return model.getFromStore(getterId, getterParams) || null
-      } else {
-        return model.getFromStore(getterId) || null
-      }
-      // return model.store.state[model.servicePath].keyedById[id.value] || null
-      // return model.getFromStore(id.value) || null
+      return model.getFromStore<M>(getterId, getterParams) || null
```
marshallswain, line 80: I believe this will cause a breaking change.  This is definitely better-written code, but I had to write it the other way due to some quirk that I couldn't figure out... and now can't remember. ;). Would you mind restoring the more-verbose option?

Other than this part, I believe this is ready to merge.

hamiltoes: Addressed in bd96d6d

marshallswain:
Merged!  Thanks @hamiltoes!  This is a huge update. Thanks for your input, as always, @J3m5!

Issue #489, opened 2020-05-23
WS: Service state ids do not include new model ids added on the client


nakedgun:
### Steps to reproduce

1. Setup feathers client + a service plugin based upon the docs: e.g. [here](https://vuex.feathersjs.com/getting-started.html#feathers-client-feathers-vuex) or [here](https://vuex.feathersjs.com/nuxt.html#working-with-auth-nuxt). 

2. Set feathers client to web sockets

3. Create a new model instance and save it:

```
const { Todo } = this.$FeathersVuex.api;
const todo = new Todo({
    text: 'Something'
});
await todo.save(); 
console.log(todo.id) // 1
```

### Expected behavior
`console.log(this.$store.state.todo.ids) // should be [1] `

### Actual behavior
`console.log(this.$store.state.todo.ids) // is actually [0]`

With **todos.ids** being out of sync with **todos.keyedById**, you get weirdness. For example, if sometime later you perform todo.find(), it will silently fail to update todo 1.

If you set the feathers client to use rest instead, it behaves as expected.

### System configuration

"feathers-client": "^2.4.0",
"feathers-hooks-common": "^4.20.7",
"feathers-vuex": "^3.9.1",

hamiltoes:
I have a hunch I broke this.

Take a look at the `updateTemp` mutation. It is moving the temp from `tempsById` to `keyedById` but it also needs to put the new id in `ids`.

nakedgun:
@hamiltoes good hunch. Adding this line seems to fix it:

```
// If an item already exists in the store from the `created` event firing
// it will be replaced here
Vue.set(state.keyedById, id, temp)
if (!state.ids.includes(id)) {
  state.ids.push(id)
}
```

marshallswain:
I'm pretty sure this can be closed, now.  Please reopen if I'm mistaken.

PR #490, opened 2020-05-24
fix: add temp's id to `ids` in `updateTemp` #489

hamiltoes:
### Summary

When promoting a temp to a "real" item, the `updateTemp` mutation migrates a temp from `tempsById` to `keyedById`. It was not also adding the new ID to the `ids` array.

This PR adds the new ID to `ids` and adds a test assertion to verify.

  - #489 
  - no

### Other Information

I changed the type of `ids` from `string[]` to `(string | number)[]` so the test could use `Array.includes` with the ID of type `number`. Ids can already be `string` or a `number`.

### DIFF
--- a/src/service-module/service-module.mutations.ts
+++ b/src/service-module/service-module.mutations.ts
@@ -151,6 +151,10 @@ export default function makeServiceMutations() {
         // If an item already exists in the store from the `created` event firing
         // it will be replaced here
         Vue.set(state.keyedById, id, temp)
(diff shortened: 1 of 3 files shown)


marshallswain:
@hamiltoes could you address the failing test?  Thanks for the PR!

hamiltoes:
@marshallswain actually the test that's failing was broke by 3.10.4 and fixed by #491 😄 is there any way to trigger Travis CI again for this PR without pushing an empty commit?

PR #491, opened 2020-05-24
fix: useGet lazy

hamiltoes:
### Summary

3.10.4 broke lazy behavior with `useGet`. This was a workaround to prevent duplicate queries due to both watchers firing. This PR reverts the commit that broke lazy behavior and combines watchers into one so the API is only hit once on initial render.

  - no
  - no

### Other Information

This PR also fixes a minor bug in the `useGet` tests where the scenario `id === null` prevents an API query. The test was also passing `lazy: true` which means the null ID was not definitely preventing the API query - it could have been prevented by `lazy: true`.


marshallswain:
Ah.  That's nicer.  Thanks!

PR #492, opened 2020-06-11
fix: addItems calls new Model calls addItem

fratzinger:
There's a loop between mutations & Model for creating models.
When you call `addItems` for completely new data, it calls `new Model` which commits `addItem` which calls `new Model` which commits `mergeInstance`


marshallswain:
Ah thanks!

PR #496, opened 2020-06-18
update 'fast-copy' to the latest version

fratzinger:
I'm implementing a localforage-blob right now. With `"fast-copy": "^2.0.5"` a blob passed to a FeathersVuexModel the returned `blob` is null because of `fastCopy` in `mergeWithAccessors`. I fixed this and _planttheidea_ merged it in `"fast-copy": "^2.1.0"`

For 99% this shouldn't change anything, but for me it's a great catch right now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,57 +13,6 @@
         "@babel/highlight": "^7.8.3"
       }
     },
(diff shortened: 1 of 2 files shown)


marshallswain:
Thanks!

Issue #504, opened 2020-07-08
FeathersVuexPagination.js:  export 'createElement' was not found in '@vue/composition-api


tvld:
### Steps to reproduce

Fresh, new install in Nuxt  2.1:
```js
// plugins/featers-client
import feathers from '@feathersjs/feathers'
import socketio from '@feathersjs/socketio-client'
import auth from '@feathersjs/authentication-client'
import io from 'socket.io-client'

import { CookieStorage } from 'cookie-storage'

let API_URL = 'https://app.bligson.com/api'

if (window.location.hostname === 'localhost') {
  API_URL = 'http://localhost:3030'
}

const socket = io(API_URL, {
  transports: ['websocket']
})

const storage = new CookieStorage()

const feathersClient = feathers()
  .configure(socketio(socket))
  .configure(auth({ storage }))

export default feathersClient

```
And then:

```js
// store/auth.js
import feathersVuex from 'feathers-vuex'
import feathersClient from '@/plugins/feathers-client'

const {
  service,
  auth,
  FeathersVuex
} = feathersVuex(feathersClient, {
  idField: '_id'
})

export default {
  service,
  auth,
  FeathersVuex
}


```
First attempt I get warning I need to install ``@vue/composition-api`` Was not clear from docs, but ``npm install save ...``.

But now I am stuck after: 
```
 ./node_modules/feathers-vuex/dist/FeathersVuexPagination.js                                                                                                                                                                                                                                                                       friendly-errors 12:20:52

"export 'createElement' was not found in '@vue/composition-api'   
```

phatj:
@tvld, the composition API entered beta status and broke a couple of things.

Use `@vue/composition-api@0.6.7` (version 0.6.7) instead:
```bash
yarn add @vue/composition-api@0.6.7

# or
npm install @vue/composition-api@0.6.7 --save
```

The `createElement` function that was imported has been aliased to just `h` as a convention (see: [here](https://vuejs.org/v2/guide/render-function.html#JSX))
> Aliasing createElement to h is a common convention you’ll see in the Vue ecosystem and is actually required for JSX. Starting with version 3.4.0 of the Babel plugin for Vue, we automatically inject const h = this.$createElement in any method and getter (not functions or arrow functions), declared in ES2015 syntax that has JSX, so you can drop the (h) parameter. With prior versions of the plugin, your app would throw an error if h was not available in the scope.

Peras795:
Ty! @phatj

[...]

PR #505, opened 2020-07-15
fix:  make createElement compatible with 1.0.0 of @vue/composition-api

phatj:
### Summary

This PR allows the usage of @vue/composition-api v1.0.0 and below as a peer dependency by dynamically using either `createElement` or `h`, whichever exists.

#504 
no

### Other Information

I am not sure if this requires a documentation update or integration testing.

### DIFF
--- a/src/FeathersVuexPagination.ts
+++ b/src/FeathersVuexPagination.ts
@@ -1,4 +1,20 @@
-import { createElement, computed, watch } from '@vue/composition-api'
+import {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
+  createElement as baseCreateElement,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
+  h,
+  computed,
+  watch
+} from '@vue/composition-api'
+import { CreateElement } from 'vue'
+
+/**
+ * Allow for usage with newer (^1.0.0) @vue/composition-api releases
+ * See: https://github.com/feathersjs-ecosystem/feathers-vuex/issues/504
+ */
+const createElement = (baseCreateElement || h) as CreateElement
 
 export default {
   name: 'FeathersVuexPagination',


J3m5:
I'm afraid that doing `import * as Composition from '@vue/composition-api'` will prevent tree-shaking with webpack or others.

phatj:
Hi @J3m5!

Good catch, I did not think about tree shaking, but I did get the idea for using a wildcard import by browsing the source of the @vue/composition-api package as it seems like wildcard exports are used [fairly](https://github.com/vuejs/composition-api/blob/master/src/index.ts) [liberally](https://github.com/vuejs/composition-api/blob/master/src/apis/index.ts).

I'm doing some more reading on tree shaking and it seems that since webpack 4, it uses a more robust analyzer for dead code elimination rather than a scan of import names.
* https://medium.com/@kz228747/tree-shaking-webpack-4-c33ce1e31e5c
* https://medium.com/@zwegrzyniak/webpack-4-1-and-es-modules-esm-dd0bd7dca4da

> An other solution would be to only support the last version of the plugin and update the version of the plugin in the package.json.

That's a great idea!  I was exploring this solution as I was inspired by Marshal's last comment in #504.

Also, there is another option:  the [createElement](https://github.com/vuejs/composition-api/blob/master/src/apis/createElement.ts) source shows that it is still exported as `createElement`, and is aliased later on [here](https://github.com/vuejs/composition-api/blob/master/src/apis/index.ts#L9).

I'm away from my machine right now, but if we just import this one function directly, feathers-vuex should be able to function normally without having to do any monkey business with wilcard exports 😄 

Let me know your thoughts! ❤️

J3m5:
> I'm super hesitant to wildcard import the entire thing. We may just need to drop support for the old syntax and do a wholesale upgrade and major version bump.

I agree

phatj:
Awesome, just updated the branch @marshallswain.  Let me know if there's anything else this PR needs.

marshallswain:
Thank you!

PR #513, opened 2020-08-06
Docs: Fix usage of the .patch() method

abrain:
The [examples for `FeathersVuexInputWrapper`](https://vuex.feathersjs.com/feathers-vuex-forms.html#feathersvuexinputwrapper) make use of the `.patch()` method of the instance object. The `data` property passed into the `save()` method contains the modifications, which can directly be used with the `.patch()` method.

However, the argument of the `.patch()` method is a params object, hence `data` must not be passed in directly, but as `{ data: data }`, or the shorthand `{ data }`.


marshallswain:
Oh good catch.  This will help avoid a lot of confusion.  Thank you!

PR #516, opened 2020-08-15
Fixed bug in require.context

engineertdog:
By not escaping the period, it has lead to multiple imports of files in my own projects, causing duplicated service errors. The proposed change ensures that `require.context` will only look at `.js` or `.ts` files.


marshallswain:
Oh man! What a find! It has been right there in front of my face for all this time and o never noticed it.  Thank you!

PR #521, opened 2020-09-01
Count - Getter/Action, Model methods & renderless data component v3.11.0

fratzinger:
### Summary

This is a replacement to #483 from @tylermmorton because there was no activity recently.

### What is this about?

Feathers.js provides a fast count query by adding `$limit: 0` to the query which is widely unknown. The proposed changes to Feathes-Vuex enable an easy and convenient way to count items on the server or store
> Pro tip: With pagination enabled, to just get the number of available records set `$limit` to `0`. This will only run a (fast) counting query against the database and return a page object with the `total` and an empty `data` array.`
https://docs.feathersjs.com/api/databases/querying.html#limit

### How is this different from #483?

**What #483 does:**
- adds the Model methods `count` and `countInStore` where the last one is a monkeypatch to the `count`-getter.
- adds tests for these two methods
- add docs for these two methods

**What this PR does:**
- add `count` getter and `count` action to service plugin
- add static `count` and `countInStore` methods to Models
- add `<FeathersVuexCount>` renderless data component for a fast way to show counts in `<template>`
- add tests for getter, action & Model-methods
- add docs for getter, action, Model-methods & renderless data components with a **v3.11.0 badge**

### But wait, there are a lot of changes in the renderless data components docs?!

As I added the third renderless data component it became difficult to remain simpleness with the existing structure:
- So now the components have individual **props** and **scope data**. It's longer but much easier to read
- changed the `slot-scope=""` on child to `v-slot=""` on parent since `slot-scope=""` is deprecated as of `vue^2.6.0` https://vuejs.org/v2/guide/components-slots.html#Scoped-Slots
- `<FeathersVuexGet>` has `params` and `fetchParams` beside `query` and `fetchQuery`. So I added it to `<FeathersVuexFind>` as well. Since the mixins and the composition API only use `params` instead of `query` it felt natural to me to **deprecate** the `query` related attributes. I documented it in the docs but it's a discussable step. The reason for `params` is it enables more flexibility, e.g. for `$populateParams`
- `<FeathersVuexCount>` only has `params` and no ``query` to keep the consistency with mixins and the composition API
- the `temps` attribute for `<FeathersVuexFind>` also gets deprecated because it gets replaced by `:params="{ query: {}, temps: true }"`

The changes got more than I thought but while programming there came more and more things along the way which felt like to they need to be pulled straight. 
@marshallswain if the `params` for `<FeathersVuexFind>` and the deprecations don't feel right to you, I will change it back. I would be happy to discuss these steps :)

Closes #483

### DIFF
--- a/src/service-module/service-module.getters.ts
+++ b/src/service-module/service-module.getters.ts
@@ -81,9 +81,7 @@ export default function makeServiceGetters() {
         throw 'params must contain a query-object'
       }
 
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      const { $sort, $limit, $skip, $select, ...cleanQuery } = params.query
-
+      const cleanQuery = _omit(params.query, FILTERS)
       params.query = cleanQuery
 
       return getters.find(state)(params).total
--- a/src/service-module/service-module.getters.ts
+++ b/src/service-module/service-module.getters.ts
@@ -19,12 +19,9 @@ const defaultOps = FILTERS.concat(OPERATORS).concat(additionalOperators)
 export default function makeServiceGetters() {
   return {
     list(state) {
-      return state.ids.map(id => state.keyedById[id])
+      return state.ids.map((id) => state.keyedById[id])
     },
-    count: state => () => {
-      return state.ids.length;
-    },
-    find: state => params => {
+    count: (state) => (params) => {
       if (isRef(params)) {
         params = params.value
       }
@@ -36,7 +33,7 @@ export default function makeServiceGetters() {
       const { paramsForServer, whitelist, keyedById } = state
       const q = _omit(params.query || {}, paramsForServer)
       const customOperators = Object.keys(q).filter(
-        k => k[0] === '$' && !defaultOps.includes(k)
+        (k) => k[0] === '$' && !defaultOps.includes(k)
       )
       const cleanQuery = _omit(q, customOperators)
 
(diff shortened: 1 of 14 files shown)


src/service-module/service-module.getters.ts
```diff
@@ -73,6 +73,9 @@ export default function makeServiceGetters() {
         data: values
       }
     },
+    count: (state, getters) => params => {
+      return getters.find(state)(params).total
```
hamiltoes, line 77:
```suggestion
      return getters.find(params).total
```

hamiltoes: I think this is being called inappropriately.

Furthermore, the `find` getter is going to do some extra stuff that is unnecessary for `count` if someone puts certain keywords in the query. I'm a fan of keeping this simple like it is because the gains are probably negligible (unless someone accidentally puts a `$sort` into a `count` on a list with thousands of items).

We could have this `count` getter delete `$sort`, `$skip`, `$limit`, and `$select` from the query just to expedite the `find` getter.

fratzinger: Thanks for your suggestion and thoughts. Your suggestion does not work. It needs to be this way.

I'm definitly on your side! I omitted the filters. Now it just passes a `cleanQuery` to the `find`-getter.

marshallswain:
@fratzinger thanks for this contribution.  I'm super grateful for this and the original idea/code contributed by @tylermmorton.  This fits perfectly into the rest of the feathers-vuex tools.

Screen Shot 2020-09-02 at 12 38 10 PM

Regarding ^^^, I've actually wanted to give the docs a more clear split like that.  It's so much easier to reference the information, so thanks for all of it.  I'm merging this as it is, now.

fratzinger:
Glad to hear this. Thanks for your quick response!
If there's anything else I can do, please feel free to ask for help. I'm happy to contribute. :)

PR #524, opened 2020-09-04
docs: seperate mixins

fratzinger:
To make the docs more consistent and as @marshallswain meant a split would be useful: Here's a start for seperating the mixins.

For now, I want to show that I'm working on it. It's not finished yet but already can be discussed.


marshallswain:
This is beautiful!  Thanks!

Issue #518, opened 2020-08-19
Question: Recommendable way to track instance status "isPatchPending, etc..."


barnacker:
### Expected behavior
I load a list of ToDo in some card grid, I delete one and I show "Deleting..." on that specific card while I'm waiting for the store to be updated and this card to disappear.

I imagined it would be nice to bind a reactive kind of status like "MyTodoInstance.isRemovePending" or something like that, like having some isRemovePending or isPatchPending but on the specific model instance that you used to call "MyTodoInstance.remove() or .save() etc..."

I was thinking of extending the current BaseModel but then I might endup with something that will break apart on your next library update so I thought asking you first haha. If that is something you are interested in I don't mind contributing.

Thx!

marshallswain:
This is a great idea. I could use this everywhere in my apps.  I would prefer to find a way to store it that doesn't end up saving to the database. Putting the extra attributes on individual model records will make for a nice API, as you've demonstrated in your comment.

hamiltoes:
Just wanted to throw in my 2c cuz I've done this in an app I was working on.

It ended up working better for me to store the individual pending status in the Vuex store on a dictionary for each feathers method. The key to each dictionary being the tempId or id of each instance. Then in my extended base model, I just had a getter, (e.g. isPatchPending) that looked up its own id in that table. This way any instance clones would also show status.

But I agree this a great feature that more people would probably use if it was included out of the box!

marshallswain:
Ah.  I'm grateful for your input @hamiltoes.  That's a really smart solution.  I hadn't thought about the clones.  If you still have access to that code it might give us a boost in getting it implemented.  If not, no worries.  I'm feeling like that's the way to built it.

barnacker:
Yes we discussed using that approach internally, the main reason being that it shouldn't break much in the upcoming updates of this package. But just in case, before coding anything, I thought of posting it here since it could be included out of the box like you mentioned.  I was imagining it done automatically when registering the feather "ToDo" service in the store.

Like you just proposed, it would automatically have an extra list like "instancesStates" the base model object would contain those getters out of the box. The .save() would set the values automatically True/False etc...

Something like that. And if the owners/contributors of this library are opened to that, I would be willing to share the final code in hopes that it be included in a version of the official library as a new feature.

hamiltoes:
Here's my lovely COVID quarantine project :-)

[...]

PR #528, opened 2020-09-06
Feature: per-instance pending status

hamiltoes:
### Summary

Adds per-instance, per-method pending status to FeathersVuex per discussion in #518 

PR includes new state, new getters, new mutations, new `BaseModel` `get`ers, and modifies the actions for `create`, `update`, `patch` and `remove` to call the new mutations.

### What Problem am I Solving?

Currently, FeathersVuex tracks `pending` status for `create`, `update`, `patch` and `remove` per service... but only in an all-or-nothing way. This means that if any model is pending, the status for that service is set. As expressed in #518, it can be useful to have a status for *this* model is patching and *that* model is removing, or similar.

In my opinion, the most useful new thing in this PR is that the `BaseModel` now has the following `get`ers:
- `isCreatePending`
- `isUpdatePending`
- `isPatchPending`
- `isRemovePending`
- `isPending` - `true` if any of the previous four statuses are `true`

### Implementation

I decided to use `Array`s of `id`s in the Vuex state for maintaining the pending status. I chose to use `Array`s instead of `Object`s because for each `id`, if a method is called multiple times, an Array can track that there are multiple requests in progress and therefore the status will not clear until the last request finishes.

I decided to make the status part of the Vuex state instead of part of the Model class to keep the status in-sync between clones and originals. E.g. when a clone is modified and `patch`ed, the original model's status indicates that a `patch` is in progress.

:heart:

*Marking this as WIP because I want to add docs and tests*

### DIFF
--- a/docs/model-classes.md
+++ b/docs/model-classes.md
@@ -339,7 +339,9 @@ const todo = new Todo({ description: 'Do something!' })
 
 The examples above show instantiating a new Model instance without an `id` field. In this case, the record is not added to the Vuex store.  If you instantiate a record **with an `id`** field, it **will** get added to the Vuex store. *Note: This field is customizable using the `idField` option for this service.*
 
-Now that we have Model instances, let's take a look at the functionality they provide. Each instance will include the following methods:
+Now that we have Model instances, let's take a look at the functionality they provide.
+
+Each instance will include the following methods:
 
 - `.save()`
 - `.create()`
@@ -348,6 +350,15 @@ Now that we have Model instances, let's take a look at the functionality they pr
 - `.clone()`
 - `.commit()`
 - `.reset()`
+
+and the following readonly attributes:
+
+- `isCreatePending` - `create` is currently pending on this model
+- `isUpdatePending` - `update` is currently pending on this model
+- `isPatchPending` - `patch` is currently pending on this model
+- `isRemovePending` - `remove` is currently pending on this model
+- `isSavePending` - Any of `create`, `update` or `patch` is currently pending on this model
+- `isPending` - Any method is currently pending on this model
 
 *Remember, if a record already has an attribute with any of these method names, it will be overwritten with the method.*
 
--- a/docs/service-plugin.md
+++ b/docs/service-plugin.md
@@ -156,7 +156,12 @@ Each service comes loaded with the following default state:
     errorOnCreate: undefined,
     errorOnUpdate: undefined,
     errorOnPatch: undefined,
-    errorOnRemove: undefined
+    errorOnRemove: undefined,
+
+    isIdCreatePending: [],
+    isIdUpdatePending: [],
+    isIdPatchPending: [],
+    isIdRemovePending: [],
   }
 ```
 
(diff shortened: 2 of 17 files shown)


hamiltoes:
Added tests for state, mutations and getters.

Still need:
- tests for actions and BaseModel
- docs

hamiltoes:
Added tests for actions and BaseModel

Still need to add documentation.

J3m5:
A Map would have been nice for that... Maybe when Vue 3 will be the main supported version....
Anyway, thanks again for your contributions, it's awesome!

fratzinger:
I like it! That's pretty awesome. I looked into it and have two annotations:
1. The `setIdPending` and `unsetIdPending` should catch a `null` id for a multi `patch`/`update`/`remove`.
2. What about also adding error-handling instance-wise? I think this could be useful and would be a consistent addition

hamiltoes:
> A Map would have been nice for that... Maybe when Vue 3 will be the main supported version....

@J3m5 could you help me understand how a map would be used here where an `object` wouldn't work?

hamiltoes:
Thanks for the feedback @fratzinger !!

> 1. The `setIdPending` and `unsetIdPending` should catch a `null` id for a multi `patch`/`update`/`remove`.

`setIdPending` does ignore IDs that aren't `typeof string` or `typeof number`. I figured `unsetIdPending` doesn't need this check because when it goes to look up the index for `null` or `undefined` it will fail.

Or do you mean the patch/update/remove actions should all handle multi and pass all IDs to `setIdPending`?

> 2. Maybe a `isSavePending` getter would add a little bit of sugar like the `save`-method 

I like this. Should just be an or of create/patch/update pending status? And should we keep `isPending` which is an or of all four? So many pending status getters! Ha

> 3. What about also adding error-handling instance-wise? I think this could be useful and would be a consistent addition

Yeah this should be easy and a straightforward addition.

fratzinger:
> `setIdPending` does ignore IDs that aren't `typeof string` or `typeof number`. I figured `unsetIdPending` doesn't need this check because when it goes to look up the index for `null` or `undefined` it will fail.

okay, perfect. As `setIdPending` has it, `unsetIdPending` doesn't need to check, I guess.

> Or do you mean the patch/update/remove actions should all handle multi and pass all IDs to `setIdPending`?

I did not think of that. I think that's what the service-wise `is${Patch|Update|Remove}Pending` is for, isn't it? :)

> I like this. Should just be an or of create/patch/update pending status? And should we keep `isPending` which is an or of all four? So many pending status getters! Ha

It's hard find the cut between to too little and too much sugar 😄 For me `isSavePending` feels natural. Also, in my applications I built a service-extension to have a `isPending` getter for all pending-states. So I would go with this as well.

> Yeah this should be easy and a straightforward addition.

Maybe we can implement this after this PR, I don't know. Let me know, if I can help :)

marshallswain:
Looks exciting, y'all! Let me know when you are ready to merge!

hamiltoes:
I've added `isSavePending` sugar and tests for it. I've added docs for all new state, mutations, getters, and BaseModel attributes.

In my opinion, per-instance error handing should be a separate PR.

Marking this ready for review.

marshallswain:
Finally getting this merged.  Thank you!  This adds huge time-saving features.

PR #529, opened 2020-09-06
Update FeathersVuexFind.ts

### LINKED ISSUE #527: Typo at FeathersVuexFind
https://github.com/feathersjs-ecosystem/feathers-vuex/blob/master/src/FeathersVuexFind.ts#L81

It should be check for `params`, not for `query`.

fratzinger:
fix for #527

It's a leftover of my last PR #521

Closes #527

### DIFF
--- a/src/FeathersVuexFind.ts
+++ b/src/FeathersVuexFind.ts
@@ -78,7 +78,7 @@ export default {
 
       params = params || { query, temps }
 
-      return query ? $store.getters[`${service}/find`](params).data : []
+      return (params && params.query) ? $store.getters[`${service}/find`](params).data : []
     },
     pagination() {
       return this.$store.state[this.service].pagination[this.qid]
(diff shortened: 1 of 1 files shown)


J3m5 (commented):
Shouldn't the query not be necessary here in case we just want to get all the items, like when we query a service without passing the params?

marshallswain:
@J3m5 is correct in this case.  This would be a breaking change.  If no query is passed, the default query object of `{}` needs to be provided.

fratzinger:
like so? Now `params` and deprecated `query` is fully optional. An empty object will be passed if falsy.

I'm confused because that wasn't the case for FeathersVuexFind before `3.12.0` neither.

marshallswain:
@fratzinger, you know what?  The latest change is actually not needed, because it's already implemented inside the `find` action, here: https://github.com/feathersjs-ecosystem/feathers-vuex/blob/master/src/service-module/service-module.actions.ts#L13

marshallswain:
@fratzinger I rarely use this module, now, so I'm going to defer to your preference on this.  Let me know when you think it's ready to merge.

PR #532, opened 2020-09-14
Update make-find-mixin.ts

fratzinger:
The `fetchParams` attribute was not supported, but the `fetchQuery` was. From time to time I like to use the `params` and `fetchParams` directly in the `makeFindMixin` options. `fetchParams` does nothing here, renaming it to `fetchQuery` directly works, but as a params-object.

Strictly speaking, this would be a breaking change, if anyone used it this way before. I can't imagine that's the case, so IMO it's a fix to be compliant with the docs

### DIFF
--- a/src/make-find-mixin.ts
+++ b/src/make-find-mixin.ts
@@ -18,7 +18,7 @@ export default function makeFindMixin(options) {
   const {
     service,
     params,
-    fetchQuery,
+    fetchParams,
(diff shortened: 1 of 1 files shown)


marshallswain:
Yes.  If this is a breaking change, it's because we're fixing a bug.  This will go out in a minor release, today.

PR #534, opened 2020-09-29
Set isDirty to false on reset()

ecirish:
From a UX perspective, this makes sense to me. If I'm resetting a clone, then there are no longer any changes present, and therefore the form is..."clean". I'm usually using a "Save" and "Reset" button which are set to `:disabled="!isValid && !isDirty"`. In this case it would be nice to reset `isDirty` so that the reset button can be disabled when no changes are present.

Do others feel the same way?


J3m5:
Yes, you are right !

marshallswain:
Thanks @ecirish!

PR #536, opened 2020-10-04
support copies in find-getter and other things

fratzinger:
### ❗ Problem:

I have a UI-table for data manipulation. I use this table somehow like the `FeathersVuexFormWrapper` but for multiple items at once. I implemented a cancel/save mechanism for the whole table and all displayed items.
So I create clones for existing items and show the clones in the table for manipulation. When I add a temp record I also concat them to the table. I implemented it by querying the original and temp records and replace the resulting items with existing clones afterwards

### ✔️ Solution:
An additional `copies` property in `params` for getters would make this shorter. I think this could become useful for someone else so here is the change for this with tests and docs. Just pass `{ query: {}, copies: true }` to activate this behaviour. By this the original and temp items get queried normally and then get replaced with the according clones. ✔️ 

### Other things:

While programming this I came accross some other issues:
1. `values = values.concat(_.values())` in the find-getter is slow. `values.push(..._.values()` is much more performant. See https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki. This optimization is only relevant for `temps: true` but since the `find` getter is used anywhere small performance changes can make a difference ✔️ 
2. The `clearCopy` mutation was not working with `Model.copiesById`. ✔️ 
3. The `createCopy` mutation was not working properly for `state.copiesById` when cloning an item twice. ✔️ 
4. I added mutation-tests for the `copy` mutations for `Model.copiesById` / `keepCopiesById: false` . But there was a problem with the getter / setter which got undefined with `mergeWithAccessor`. I could not solve this problem. 😕 
5. I wrote tests for `it.skip('instance.remove removes cloned records from the store', function () {})` and `it.skip('instance.remove removes cloned records from the Model.copiesById', function () {})` but kept them as `it.skip` because I don't know whether you want to implement this, or not @marshallswain. See the discussion below: ❓ 

### ❓ Discussion:

There are some things I'm not sure how it should be. At the moment `copiesById` can get poluted because it never releases items (just by `clearCopy` which never gets called):
a. `removeItem` / `Model.remove` should delete the item from `copiesById`, as it is suggested in the tests? 💯 
b. `resetCopy` / `Model.reset` should delete the item from `copiesById`, shouldn't it? 💯 
c. ``commitCopy` / `Model.commit` should also delete the item from `copiesById`, right? I'm not that sure about this one. 😕 
d. `clearAll` mutation should also clear `copiesById`, shouldn't it? 💯 

@marshallswain For the last four questions I need your thoughts, please. As always I'm happy to contribute. Just a 'yes' or 'no' would be enough and I will implement them. 👍

### DIFF
--- a/docs/composition-api.md
+++ b/docs/composition-api.md
@@ -119,6 +119,7 @@ And here's a look at each individual property:
   - Set `params.paginate` to `true` to turn off realtime updates for the results and defer pagination to the API server.  Pagination works the same as for the [makeFindMixin pagination](./mixins.html#pagination-with-fall-through-cacheing).
   - Set `params.debounce` to an integer and the API requests will automatically be debounced by that many milliseconds.  For example, setting `debounce: 1000` will assure that the API request will be made at most every 1 second.
   - Set `params.temps` to `true` to include temporary (local-only) items in the results. Temporary records are instances that have been created but not yet saved to the database.
+  - Set `params.copies` to `true` to include cloned items in the results. The queried items get replaced with the corresponding copies from `copiesById`
 - `fetchParams` This is a separate set of params that, when provided, will become the params sent to the API server.  The `params` will then only be used to query data from the local data store.
   - Explicitly returning `null` will prevent an API request from being made.
 - `queryWhen` must be a `computed` property which returns a `boolean`. It provides a logical separation for preventing API requests *outside* of the `params`.
--- a/docs/service-plugin.md
+++ b/docs/service-plugin.md
@@ -199,7 +199,10 @@ Service modules include the following getters:
 
 - `list {Array}` - an array of items. The array form of `keyedById`  Read only.
 - `find(params) {Function}` - a helper function that allows you to use the [Feathers Adapter Common API](https://docs.feathersjs.com/api/databases/common) and [Query API](https://docs.feathersjs.com/api/databases/querying) to pull data from the store.  This allows you to treat the store just like a local Feathers database adapter (but without hooks).
-  - `params {Object}` - an object with a `query` object and optional `paginate` and `temps` boolean properties. The `query` is in the FeathersJS query format. You can set `params.paginate` to `false` to disable pagination for a single request.
+  - `params {Object}` - an object with a `query` object and optional properties. You can set the following  properties:
+    - `params.query {Boolean}` - The `query` is in the FeathersJS query format.
+    - `params.temps {Boolean}` - **Default:** `false` - if `true` also consider temporary records from `tempsById`
+    - `params.copies {Boolean}` - **Default:** `false` - if `true`: first search for the regular records and then replace the records with the related copies from `copiesById`
 - `count(params) {Function}` - a helper function that counts items in the store matching the provided query in the params and returns this number <Badge text="3.12.0+" />
   - `params {Object}` - an object with a `query` object and an optional `temps` boolean property.
 - `get(id[, params]) {Function}` - a function that allows you to query the store for a single item, by id.  It works the same way as `get` requests in Feathers database adapters.
--- a/src/service-module/make-base-model.ts
+++ b/src/service-module/make-base-model.ts
@@ -79,7 +79,8 @@ export default function makeBaseModel(options: FeathersVuexOptions) {
     public static serverAlias: string = options.serverAlias
 
     public static readonly models = globalModels as GlobalModels // Can access other Models here
-    public static copiesById: {
+
+    public static readonly copiesById: {
       [key: string]: Model | undefined
       [key: number]: Model | undefined
     } = {}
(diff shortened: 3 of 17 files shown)


J3m5:
Just my 2 cents:
a. Yes
b. Not sure, would it mean that we need to re-clone the item ?
c. Yes
c bis. Yes

I noticed this behavior when I re-implemented the VuexFormWrapper for a specific need and manually added the copy deletion.

About the copies param, in fact, it already exist, it's [there](https://github.com/feathersjs-ecosystem/feathers-vuex/blob/7b4de7dbdc52919a29c528785aab60d73f0c4e6d/src/service-module/service-module.getters.ts#L31) 

```js
// Set params.temps to true to include the tempsById records
params.temps = params.hasOwnProperty('temps') ? params.temps : false
```
but it's called `temps` and is [mentioned in the docs here](https://vuex.feathersjs.com/3.0-major-release.html#renderless-data-components-query-fetchquery-and-temps-badge-text-deprecated-type-warning) but maybe you already saw it.

What I'm not sure there is why the original items are concatenated with the copies ?
Shouldn't they replace the original ones ?

fratzinger:
Thanks for your thoughts 👍 

How did you add the custom copy deletion? If you don't mind sharing maybe we can bake that into `FeathersVuex` to make it even better? I thought of calling the `clearCopy` mutation in a `beforeDestroy` / `beforeUnmount` hook of the renderless data component.

I think you misunderstood the `temps` param:
- `temps` -> `state.tempsById` store new records that have not been send to the server and therefore don't have an `id`
- `copies` or also `clones` -> `state.copiesById` / `Model.copiesById` store records cloned from original records (`keyedById` or `tempsById`)

So they basically are two seperate things. Temps for complete new records. Copies for a tempory discard/save mechanism.

J3m5:
Ah yes, definitely, now that you say it I remember that I was confused at first and then understood the difference.
I just forgot it in the meantime... ^^
So yes it totally make sense.

For the copy deletion, I did it manually by deleting the id in copiesById on the item constructor like this

```js
deleteCopy() {
  const { idField, copiesById } = this.item.constructor;
  const id = this.clone[idField];
  Reflect.deleteProperty(copiesById, id);
},
```

marshallswain:
@fratzinger, I'm going to start by answering the last four questions, then I'll go back and review the rest.

> a. removeItem / Model.remove should delete the item from copiesById, as it is suggested in the tests? 💯

**Yes**

> b. resetCopy / Model.reset should delete the item from copiesById, shouldn't it? 💯

**No**  Reset copy should basically do the same thing as re-cloning.  It resets the copy back to match the original record.

> c. ``commitCopy/Model.commit` should also delete the item from `copiesById`, right? I'm not that sure about this one.

**No**  My most common pattern is to work from a copy and send commits/saves back to the original/API server.  This would break literally every form I've ever written.  :)

> d. `clearAll` mutation should also clear `copiesById`, shouldn't it? 💯

**Yes**

marshallswain:
@fratzinger this is an excellent idea.  I'm curious to try it out and compare it to my current workflow, which uses a new `handleClones` utility that wrote a couple of months ago.  I think it's time to bring it into feathers-vuex and share it.

#4 is interesting.  `mergeWithAccessors` is quite a pain to work with, so I'm not surprised.  Please at least leave a skipped test in place for that.

Overall it sounds great.  Let me know when you'd like it merged and I'll do a release.

fratzinger:
Wow, this was more to do than I expected.

### What was added to this PR since the last discussion?
- `clearAll` also clears the `copiesById`
- `Model.remove` also removes the corresponding `copy`
- mutation `removeItem` also removes the corresponding `copy`
- mutation `removeItems` also removes the corresponding `copies`
- added tests for these

### But wait, why are there unrelated changes in the tests?
Handling the `state/Model`.`copiesById` thing in the tests was more difficult than I expected. It said `Cannot get 'copiesById' of undefined`. After digging and digging deeper I found out it has to do with the 1on1 relation between `service-path` and `service/Model`. Since most of the services are called `todos` in the several tests-files it's not hardly debugable. The tests in `model.methods.tests.js` called states from `make-service-plugin.test.js` and so on. So I tried to straighten that out:
- custom `server-alias` per file
- custom `service-path` per file
I recommand doing this from now on, even if it was not a problem before.

It's ready from my perspective. Let me know @marshallswain if anything is missing or incorrect.

marshallswain:
@fratzinger would you mind fixing the conflicts.  I'm not sure why it won't let me do it right here in the Github UI.

marshallswain:
Oh, also I forgot to say that this pull request is beautiful and I really appreciate how thorough you've been.  The changes to the tests look great.

fratzinger:
conflicts resolved.

Bizarrely there were new errors in the debounce events tests, that drive me nuts. I guess the `performance.now()` provides not the best developing experience.
I don't know how the `Uncaught Unavailable: connect ECONNREFUSED 127.0.0.1:3030` can be resolved.

Just one last thing came into my mind regarding the initial intention of this PR: The `copies: true` param just adds copies from `keyedById` but no copies from `tempsById` by now. What do you think @marshallswain? Should I change this to also include copies of temp records? I'm confused about cloning a temp anyway.

marshallswain:
@fratzinger. Thank you for this entire thing!  I'll create an issue to support copies from tempsById, too.  This is a valid use case.

marshallswain:
@fratzinger here's the issue.  It's low priority for right now, since temps are less commonly used: https://github.com/feathersjs-ecosystem/feathers-vuex/issues/554

fratzinger:
perfect. Thanks :)

marshallswain:
Released as `feathers-vuex@3.15.0`.  Thank you!

PR #537, opened 2020-10-05
Debounce events

### LINKED ISSUE #533: [Question] Optimizing for bulk patch responses and update cycle on the front-end
Hey - What is the recommended way of handling bulk updates on the client-side?

Right now, when I issue a bulk patch request (say adding a new tag to a lot of items) - The request gets executed properly but the update responses from the server make things extremely slow on the client. 

So typically, I would issue a request which tells the API to add a tag on items with X,Y,Z ids. The API executes the bulk patch and returns the updated objects of X, Y and Z.

After digging around a bit, I see that the updates get processed one by one as the updated objects are received from the API so, from the front-end perspective, there is not much difference in terms of how the bulk patch works vs issuing individual patch requests for all the items. Is there something that is wrong in my setup or is this the expected behavior? 

What would be the right way to handle this without freezing the app when you're updating say 100 or more records?

Thanks!

fratzinger:
WIP

Closes #533

### DIFF
--- a/src/service-module/make-service-plugin.ts
+++ b/src/service-module/make-service-plugin.ts
@@ -6,8 +6,14 @@ eslint
 import { FeathersVuexOptions, MakeServicePluginOptions } from './types'
 import makeServiceModule from './make-service-module'
 import { globalModels, prepareAddModel } from './global-models'
-import { makeNamespace, getServicePath, assignIfNotPresent } from '../utils'
+import {
+  makeNamespace,
+  getServicePath,
+  assignIfNotPresent,
+  getId
+} from '../utils'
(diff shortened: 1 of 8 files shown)


marshallswain:
@fratzinger this is great.  I pulled your edits, made some mostly cosmetic changes, and pushed the code to a new branch, here: https://github.com/feathersjs-ecosystem/feathers-vuex/tree/pullrequests/fratzinger/debounce-events

The only big change is that I moved the event handling into its own file.  I've given you write access to the feathers-vuex repo, so you can make edits to that branch for tests and docs.  Could you edit your pull request to source that branch, instead?  Thanks!

Issue #558, opened 2020-12-08
adding debounce to watch prop of renderless data components


jd1378:
I'm creating a pull request soon for this, just wanted to inform you and also ask if should I go for `debounce` instead of `debounceWatch` name for the new prop ? I'm going for debounce myself because I think the component doesn't fetch again without watch prop, right ?
new prop is going to accept number in milliseconds and use lodash's debounce
It's pretty much straightforward. but also I didn't see any test for the watch prop of renderless component, so I didn't write any tests either
any thoughts will be appreciated.

jd1378:
actually, I go for `debounceWatch` , since it may confuse the behaviour of normal usage without the `watch` property

PR #559, opened 2020-12-08
add debounceWatch prop to renderless components

jd1378:
solves #558


marshallswain:
Hey could you please reopen this against the Vue-demi branch?  I'm about to publish 4.0 and this will get lost in the transition.  You will probably need to start fresh from that branch.

PR #562, opened 2020-12-14
Replace all hyphen characters

sebastien-prudhomme:
### Summary

When using a Feathers service name with more than one hyphen character, feathers-vuex mixins no longer works when generating computed properties names and so on.

The current code only replace the first hyphen character.

This is a fix to replace all hyphen characters.


marshallswain:
Ah, good catch!  Could you please reopen this against the vue-demi branch?  That will speed up the process of getting the Vue2/Vue3 version out the door, hopefully this week.

Issue #565, opened 2020-12-24
[BUG] Cant update model with id 0


DaniGTA:
### Steps to reproduce

Have a database with ids like 0, 1, 2, 3, 4... and so on
create a model with id 0 and try to update it

### Expected behavior
Should call `_dispatch.call(this.constructor, 'update', [id, this, params])`

### Actual behavior
Calls: 
```
const error = new Error(
  `Missing ${idField} property. You must create the data before you can update with this data`
)
return Promise.reject(error)
```

### Caused by:

make-base-model.js

```
330: if (!id) {

}
```

when id is 0 it cant update. It must check if it is undefined or null.

PR #566, opened 2020-12-24
FIX Can't call update when id is 0

DaniGTA:
Fix for Databases that use numbers as ID's and model's with number id 0 should be able to call the function 'update'.

#565


marshallswain:
Good catch. Thank you,  I've made one tiny request.

marshallswain:
Thank you!

PR #577, opened 2021-02-25
vue-demi: feat: static model methods `create`, `update`, `patch`, `remove` + docs

fratzinger:
As discussed on slack. 

Weirdly the VSCode takes the right declarations but the wrong jsdoc-comment (from instance methods):
image

I don't know what to do about it. If someone got a pointer, I would happily change that, but otherwise I think this can get added as is.

### DIFF
--- a/docs/model-classes.md
+++ b/docs/model-classes.md
@@ -184,7 +184,7 @@ created () {
 }
 ```
 
-### get(id, params)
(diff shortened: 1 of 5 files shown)


marshallswain:
Thanks!

fratzinger:
as always happy to help! :)

Issue #590, opened 2021-04-07
Add support for the `disablePagination` hook in `feathers-hooks-common`


joshuaja:
### Steps to reproduce

`feathers-hooks-common` has a [disablePagination](https://hooks-common.feathersjs.com/hooks.html#disablepagination) hook that allows us to pass in `$limit: -1` to disable pagination for a Feathers service.

When passing `$limit: -1` to `useFind` params in `feathers-vuex`, we have to use both `params` and `fetchParams` as a workaround since `feathers-vuex` doesn't see `$limit: -1` as turning off pagination.

### Expected behavior
If `$limit: -1` is passed to `useFind` `params` I would expect for it to add `paginate: false` to the params and `$limit: -1` to the fetchParams in the background.

### Actual behavior
`feathers-vuex` doesn't turn off pagination if `$limit: -1` is passed to `params`.

### System configuration

**Module versions** (especially the part that's not working):
- `"feathers-vuex": "^3.15.0"`
- `"vuex": "^3.6.2"`

**NodeJS version**:
- `"node": "14.4.0"`
- `"npm": ">= 6.14.5"`

**Operating System**:
- Linux/Docker

**Browser Version**:
- Latest Chrome

**Module Loader**:
- Vue CLI

fratzinger:
That's something I have to do quite often. FeathersVuex takes `$limit: -1` as `$limit: 1` and just returns one item.
@marshallswain: is this something that can work natively without any breaking changes for anyone not working with `disablePagination`? So should this just work by default or should it be an option (e.g.: `addSupportForLimitMinusOne`)?

As I use it quite often, I would like to add this feature.

J3m5:
Too me it seems that we just need to handle the -1 value specifically in the find getter.
Currently, it is just passed as the last parameter of the slice method [there](https://github.com/feathersjs-ecosystem/feathers-vuex/blob/master/src/service-module/service-module.getters.ts#L88).
We could detect the value and act accordingly by not slicing the items and returning just the items, not the paginated result.

Yavorss:
> To me, it seems that we just need to handle the -1 value specifically in the find getter.
> Currently, it is just passed as the last parameter of the slice method [there](https://github.com/feathersjs-ecosystem/feathers-vuex/blob/master/src/service-module/service-module.getters.ts#L88).
> We could detect the value and act accordingly by not slicing the items and returning just the items, not the paginated result.

[...]

PR #588, opened 2021-03-16
service.module.getters refactor

J3m5:
- unref instead of isRef
- skip/limit optimization
- replace _.values
- get copies with map instead of forEach
- avoid params mutation
- remove lodash get

Related to #587

PR #607, opened 2021-07-02
paramsForServer as key plus value or result of a function

Yavorss:
### Summary
This a solution to the issue with `-1` from `feathers-hooks-common` mentioned here https://github.com/feathersjs-ecosystem/feathers-vuex/issues/590. The solution is simple. Instead of filtering only by parameter, we could filter by parameter with a specific value. For instance, `paramsForServer: [['$limit', '-1']]` or `paramsForServer: [['$limit', value => '-1']]`.

### Other Information
 If that looks good for the team, we could update the documentation and link that PR to here as well.

### DIFF
--- a/src/service-module/service-module.getters.ts
+++ b/src/service-module/service-module.getters.ts
@@ -48,8 +48,29 @@ export default function makeServiceGetters() {
       // Set params.copies to true to include the copiesById records
       params.copies = params.hasOwnProperty('copies') ? params.copies : false
 
-      const { paramsForServer, whitelist, keyedById } = state
-      const q = _omit(params.query || {}, paramsForServer)
+      const { paramsForServer = [], whitelist, keyedById } = state
+
+      const paramsForServerByValue = paramsForServer.filter(el => Array.isArray(el))
+
+      const q = paramsForServerByValue.reduce(
+        (acc, [key, filter]) => {
+          if (!acc[key]) return acc
+
+          if (
+            ((typeof filter === 'string' || typeof filter === 'number') &&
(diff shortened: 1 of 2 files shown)


J3m5:
Please, take a look at my comment [there](https://github.com/feathersjs-ecosystem/feathers-vuex/issues/590#issuecomment-872939295)

joshuaja:
@J3m5 Allowing a filtering function within the`paramsForServer` solves the `$limit: -1` issue and also adds some needed functionality to the `paramsForServer` filter.

That would be a better and more straightforward solution over hard-coding the `-1` in multiple places within the code.

J3m5:
Ok I get it.

There is one small problem however.
This MR will conflict [with a previous one](https://github.com/feathersjs-ecosystem/feathers-vuex/pull/588)

And I think [the added code](https://github.com/feathersjs-ecosystem/feathers-vuex/pull/607/files#diff-31460bc3932b76faa5853e7340071c1e6a04d435cf964ec16702803cf0a26e09R53-R77) could be isolated in a separate function.

marshallswain:
This is a good idea.  Can you double check the conflicts from the other PRs merged?

Yavorss:
> This is a good idea. Can you double check the conflicts from the other PRs merged?

Will update it these days or even today.

fratzinger:
Pretty nice! I also thought about this before.

A mention about this in the docs would be pretty neat! @Yavorss could you add this to the docs?

marshallswain:
@Yavorss were you able to make any additional progress on this.  I should have some time soon to take a look, if not.

